PR #28, opened 2015-06-25
GApps for Stock merge

rapperskull:
The "fornexus" variant, aka Open GApps for Stock is probably ready. I flashed it successfully on my Nexus 5 and only need to be tested by more people.


mfonville:
What will actually happen if somebody will flash the fornexus with the inclusion and exclusion of certain applications? Will those then e.g. be removed from their device?

rapperskull:
I understand, but having a folder called . in the root of the tar file is annoying.
We can use:
cd /tmp;
tar -cz -f "$log_folder/open_gapps_debug_logs.tar.gz" logs/*;
So that the wildcard problem is solved and there's no . folder in the file.

As for as the fornexus package, it's exactly like the Stock package, with the exception that will flash on top of other GApps packages and that removes all the old Chrome libraries from /system/lib. The new removals for files found in some ROMs have been added to the general code: they will simply never be used.

mfonville:
Yup, that path is better than using the dot.

My question was a bit focused on the hypothetical case, that since the Nexus has a lot of applications pre-installed, what would happen if one excludes an app. But then I assume that any excluded app will thus be deleted from /system/.

rapperskull:
Yes, the gapps-remove.txt is the same, so all the GApps in the package are removed. I too had to exclude some apps on my Nexus 5 because of the space.

mfonville:
I merged all the work into master

mfonville:
Btw, since the work is now in master, the buildbot will build it tomorrow automatically. So it will be uploaded under the 'fornexus' keyword on the GitHub Releases page.

rapperskull:
Do you have to update the code of the gh-pages branch or is all automatic?

mfonville:
I will update tomorrow the gh-pages branch, when the download will be online.
I have created this wiki page: https://github.com/opengapps/opengapps/wiki/For-Nexus-Package probably it is best if you write there some general information and link to your XDA thread, and then I will link to that article from the Open GApps homepage.

rapperskull:
I'll edit the Wiki page ASAP. Tomorrow I'll test the build and if all is fine I'll update my XDA thread.

mfonville:
Btw, is your script compatible with arm64 Nexus images?

rapperskull:
That should work on arm64 and even x86 devices, from this point of view there's no difference with the stock package.

mfonville:
Ok, great. Just that on arm64 there is a 'lib64' folder, so didn't know if it would account for the existance of it :).

rapperskull:
The lib64 is not a problem, however I downloaded the Nexus 9 factory image and found out that there's no libchrome at all, both in /system/lib and in /system/lib64. I made a commit to also remove /system/lib64/libchrome*.so if they're found on some devices.

PR #32, opened 2015-06-27
Change domain to opengapps.cf

### LINKED ISSUE #31: domain and changing projectpage to organizationpage
One thing I noticed is how long the URL for the download site is... We could register something like `opengapps.cf` for free and then forward the GitHub page to it.

Three things about this:

1. Old domain would still redirect!
2. Does require DNS but that can be free also (aka [ClouDNS](http://cloudns.net))
3. Not as memorable as a `.com` but would still be a lot shorter and easier to remember than opengapps.github.io/opengapps

Any thoughts guys? @opengapps/general 

Task List
=======
- [x] Get Free Domain
- [x] Forward free domain to GitHub Pages
- [x] Look into a "Paid" domain like ~~`.com`~~`.org`
- [x] Forward "Paid" domain to GitHub Pages
- [x] Update README to point to new domain
- [x] Move Project page to Organization Page
- [x] Redirect `http://opengapps.org/opengapps` to `http://opengapps.org`
- [ ] Close 'OLD' `gh-pages` branch -> there are too many links to this old URL. We cannot abandon it for a long time.

bgiesing:
Fixes #31

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,0 +1,1 @@
+opengapps.cf


mfonville:
I will put this on hold, I rather want to use opengapps.org, which I will set up tomorrow.

PR #33, opened 2015-06-27
Change README to point to opengapps.org

bgiesing:
Forgot to point README! (old domain would still work though, but consistency is nice!)

Part of #31 and #32


mfonville:
I will put this on hold, I rather want to use opengapps.org, which I will set up tomorrow.

Issue #48, opened 2015-07-06
ShellCheck linter issues


ghost:
I ran ShellCheck 0.3.8 against 133c884 and figured I'd post the output here in case anyone was interested in going through it. If no one is interested, I'll see what I can do and then open a PR for review.

Going by a quick glance, it looks like there's possibly quite a few false positives, so I've split up the output into errors, warnings and notes so that it's easier to go though all of them.

[shellcheck-plaintext](http://hastebin.com/oferaxagew)
[shellcheck-errors](http://hastebin.com/uzorisakaw)
[shellcheck-warnings](http://hastebin.com/ulikanelux)
[shellcheck-notes](http://hastebin.com/fupubimuxe)

shellcheck --format=gcc --shell=sh

```
add_sourceapp.sh:33:10: note: Use $(..) instead of legacy `..`. [SC2006]
add_sourceapp.sh:33:16: note: Double quote to prevent globbing and word splitting. [SC2086]
add_sourceapp.sh:58:85: note: tr replaces sets of chars, not words (mentioned due to duplicates). [SC2020]
add_sourceapp.sh:59:86: note: tr replaces sets of chars, not words (mentioned due to duplicates). [SC2020]
add_sourceapp.sh:60:66: note: tr replaces sets of chars, not words (mentioned due to duplicates). [SC2020]
add_sourceapp.sh:73:12: note: Use $(..) instead of legacy `..`. [SC2006]
add_sourceapp.sh:75:19: note: Use $(..) instead of legacy `..`. [SC2006]
add_sourceapp.sh:89:12: note: Use $(..) instead of legacy `..`. [SC2006]
add_sourceapp.sh:89:23: note: $/${} is unnecessary on arithmetic variables. [SC2004]
add_sourceapp.sh:104:19: note: Double quote to prevent globbing and word splitting. [SC2086]
add_sourceapp.sh:105:17: warning: The surrounding quotes actually unquote this. Remove or escape them. [SC2027]
add_sourceapp.sh:105:17: note: Double quote to prevent globbing and word splitting. [SC2086]
add_sourceapp.sh:105:42: warning: The surrounding quotes actually unquote this. Remove or escape them. [SC2027]
add_sourceapp.sh:105:42: note: Double quote to prevent globbing and word splitting. [SC2086]
add_sourceapp.sh:105:71: warning: The surrounding quotes actually unquote this. Remove or escape them. [SC2027]
[...]

PR #52, opened 2015-07-07
Replace echo -n usage with printf

Bracket-:
**UNTESTED**
Please test and review, I have not tested these changes.

echo -n is implementation-defined and therefor not portable

Never use echo flags unless your script specifies a specific shell to be used (e.g. #!/bin/bash) to ensure consistent flag behavior. Please see this page for more information: [pubs.opengroup.org/onlinepubs/9699919799/utilities/echo.html](pubs.opengroup.org/onlinepubs/9699919799/utilities/echo.html). Specifically [OPERANDS](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/echo.html#tag_20_37_05), [APPLICATION USAGE](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/echo.html#tag_20_37_16) and [RATIONALE](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/echo.html#tag_20_37_18).

Partially fixes issue #48

### DIFF
--- a/scripts/inc.buildhelper.sh
+++ b/scripts/inc.buildhelper.sh
@@ -78,7 +78,7 @@ buildapp(){
 			if [ "$versionname" = "$baseversionname" ]; then
 				density=$(basename "$(dirname "$dpivariant")")
 				buildapk "$dpivariant" "$ziplocation/$density/$targetlocation"
-				echo -n " $density"
+				printf " %s" "$density"
 				echo "$ziplocation/$density/" >> "$build/app_densities.txt"
 			fi
 		done
--- a/scripts/inc.installdata.sh
+++ b/scripts/inc.installdata.sh
@@ -41,7 +41,7 @@ $gapps_remove"
 			fi
 		done
 	done
-	echo -n "$gapps_remove" | sort > "$build/gapps-remove.txt" # Use the -n switch to avoid a newline at the beginning of the file.
+	printf "%s" "$gapps_remove" | sort > "$build/gapps-remove.txt"
 }
 makeinstallerdata(){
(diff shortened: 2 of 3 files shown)


scripts/inc.buildhelper.sh
```diff
@@ -78,7 +78,7 @@ buildapp(){
 			if [ "$versionname" = "$baseversionname" ]; then
 				density=$(basename "$(dirname "$dpivariant")")
 				buildapk "$dpivariant" "$ziplocation/$density/$targetlocation"
-				echo -n " $density"
```
Bracket-: Was the space here intentional?

scripts/inc.installdata.sh
```diff
@@ -41,7 +41,7 @@ $gapps_remove"
 			fi
 		done
 	done
-	echo -n "$gapps_remove" | sort > "$build/gapps-remove.txt" # Use the -n switch to avoid a newline at the beginning of the file.
+	printf "%s" "$gapps_remove" | sort > "$build/gapps-remove.txt" # Use the -n switch to avoid a newline at the beginning of the file.
```
mfonville: You can then remove the comment.

scripts/inc.buildhelper.sh
mfonville: Yes, I believe so. To not write the dpi attached to the versionnumber or a previous dpi.

scripts/inc.installdata.sh
Bracket-: :thumbsup: Thanks for the catch!

PR #54, opened 2015-07-07
Fix and properly align app_sizes.txt

Bracket-:
\t was being output literally and would not have aligned the columns anyway. Use printf to properly align the columns.  
Before: [http://hastebin.com/ametusogor.txt](http://hastebin.com/ametusogor.txt)
After: [http://hastebin.com/imogebuwip.txt](http://hastebin.com/imogebuwip.txt)  
The printf relies on googleonetimeinitializer and 400-480-560-640 being the largest strings in their columns. It's not worth complicating things by adding width detection for the columns just for cosmetics.  
If you're not happy with using hardcoded values, we can always switch to plain printf with unaligned tabs or echo with spaces since it will be unaligned anyway.  
Partially fixes issue #48

### DIFF
--- a/scripts/inc.packagetarget.sh
+++ b/scripts/inc.packagetarget.sh
@@ -62,7 +62,7 @@ createzip() {
 		for f in $(ls); do # ls is safe here because there are no directories with spaces
 			for g in $(ls "$f"); do
 				foldersize="$(du -ck "$f/$g/" | tail -n1 | awk '{ print $1 }')"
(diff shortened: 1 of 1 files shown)


Bracket-:
Can someone confirm get_appsize() of scripts/inc.updatebinary.sh still works as expected? I'm not familiar with awk. Just need to confirm the new spacing/character doesn't break the awk command.

Issue #56, opened 2015-07-09
AROMA is broken


mfonville:
Aroma build is broken since we have implemented the compression method.

Suspicion is that there might be lack of RAM because AROMA also needs some memory, while extracting whole gmscore to /tmp/ (which is also fully in RAM) takes e.g. 150MiB.

so, you get a big pressure on RAM, which might even be the problem for some people who are having trouble installing the current 'stock' package:
* Aroma takes RAM
* TWRP takes RAM
* a .xz file gets unzipped into /tmp/ which takes RAM
* the .xz gets further extracted (150MiB) into /tmp/ which takes RAM
* the extraction process of the xz-file also takes RAM

rapperskull:
Do we have logs to analyze?
One thing we can think of, if the memory is the real problem, is to only generate the config file with the AROMA zip and then let the user flash the stock package. It's however the worst case scenario.

kronflux:
Aroma restarts TWRP halfway through flash. Here's all I have for a log:

I:Set page: 'install'
I:Set page: 'flash_confirm'
I:Set page: 'flash_zip'
I:operation_start: 'Flashing'
Installing '/sdcard/flashable_zips/open_gapps-arm-5.1-aroma-20150710.zip'...
Checking for MD5 file...
Skipping MD5 check: no MD5 file found
I:Zip does not contain SELinux file_contexts file in its root.
I:Legacy property environment initialized.
aroma/s: Initializing
Starting AROMA Installer version 2.70RC2
     (c) 2013 by amarullz xda-developers

Starting AROMA Installer version 2.70RC2
(c) 2013 by amarullz xda-developers

aroma/s: Mute Parent
aroma/s: Saving Arguments
aroma/s: Open Archive
aroma/s: Initializing Resource
aroma/s: Opening Freetype
aroma/s: INDR Input Device: hs_detect (event3) - Class : 1
aroma/s:   VKN : 0, CALIB : (0,0,0,0)
aroma/s: INDR Input Device: touch_dev (event1) - Class : 6
aroma/s:   VKN : 0, CALIB : (0,1079,0,1919)
aroma/s: INDR Input Device: qpnp_pon (event0) - Class : 1
aroma/s:   VKN : 0, CALIB : (0,0,0,0)
aroma/s: INDR Input Device: gpio-keys (event2) - Class : 1
aroma/s:   VKN : 0, CALIB : (0,0,0,0)
aroma/s: Starting Interface
[...]

PR #59, opened 2015-07-11
Move the tar.xz extraction to /cache/Open-GApps/ to prevent RAM trouble

mfonville:
I hope that this change should fix issue #56 and the other problems people experience with e.g. a themed TWRP, by reducing the amount of RAM that is used. But it needs to be tested and I would like it if @rapperskull could also review this change.


rapperskull:
The change looks legit, but I'm not at home and I can't test it. Another improvement we could make is to check for free space in /cache and delete the content if there's not enough.

mfonville:
It didn't solve the problem :-/ so we abandon this branch/pull

PR #63, opened 2015-07-14
Update add_sourceapp.sh for incorrect end of case sections

NicholasBuse:
Update add_sourceapp.sh for incorrect end of case sections


mfonville:
thanks. You are also doing great work with all those updated packages :-)

PR #75, opened 2015-07-23
Add automated upload script

### LINKED ISSUE #73: Automated Upload Script
Maybe we can add a simple bash script (maybe `./upload.sh`?) that would automatically cd into the source directory, add the files, commit with a common format, and push.
```
cd sources/$arch &&
git add . &&
git commit -am'$AppName $VersionName ($DPI)' &&
git push origin HEAD:master &&
cd ../../
```

To explain:

1. The `cd` command would cd into the appropriate source dir.
2. Add to Staged
3. Commit with $AppName being either the keyword, package name, or the Play Store name; $VersionName and $DPI coming from the `./add_sourceapp.sh` command variables.
4. This is interesting, but when you clone and run `./download-sources.sh`, it doesn't set the branch of the modules to master, so this version of the command would take what you just added (HEAD) and push it to origin/master.
5. Take you back to the main directory.

If we have multiple files added at the same time, then we could do two things:

- Add to one commit (with the name being altered to show `$AppName $VersionName ($DPI) and more` (Or something similar)
- Add to one commit and put all the `$AppName $VersionName ($DPI)` on a line each.
- Make seperate commits for each APK of different name (aka multiple GMS versions would still be one commit with the name having multiple DPIs, but Play Store and Chrome would still be seperate commits)

Any thoughts or anybody want to implement?

bgiesing:
Fixes #73 

@mfonville I tried following the suggestions, but if I did anything with $1 or moving $apk to the loop, it would not get the properties.

I did as much as I can. At least it works and if you want to modify it a bit, I wouldn't mind.

(Sorry if I sound frustrated... I am normally better but I haven't had sleep in 3 days.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,10 +34,7 @@ One can update the sources used with the command:
 ```
 ./add_sourceapp.sh /path/to/the/file/you/want/to/add.apk [/even/more/files.apk...]
 ```
-For contributors, updated sources can be uploaded by browsing to one of the sources-subdirectories and pushing it to the repository:
+For contributors, updated sources can be uploaded by using this command:
 ```
-cd sources/architecture
-git add *
-git commit -a
-git push
+./upload.sh
 ```
--- a/upload_sources.sh
+++ b/upload_sources.sh
@@ -1,0 +1,40 @@
+#!/bin/sh
+#This file is part of The Open GApps script of @mfonville.
+#
+#    The Open GApps scripts are free software: you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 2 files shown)


bgiesing:
@mfonville I tried following the suggestions, but if I did anything with $1 or moving $apk to the loop, it would not get the properties.

I did as much as I can. At least it works and if you want to modify it a bit, I wouldn't mind.

bgiesing:
@mfonville I couldn't get the git rm to work either so... It needs to be added still.

mfonville:
I had to change still some things, check out the uploadscript branch.

PR #108, opened 2015-09-09
Fix 

mc1100:
Tested, both certificate output formats work (with the new line terminator after "Serial Number:" field and with the serial number on the same line). Unfortunately this adds dependency on the "bc" tool as I do not know any other reliable way to convert 64-bit hex integers directly in awk.


mc1100:
Only the last commit is relevant.

mc1100:
BTW. is there a way to include only specific commit(s) in the pull request?

mfonville:
Yes, it should be, by redefining the pull request. But I have no idea how, I never had to do any PR. Maybe one of the other @opengapps/packagers knows.

PR #112, opened 2015-09-20
Support forceclean and skipkeybdlib tokens in gapps-config and tweak keyboard treatment

eugenesan:
For example gapps-config can now contain:
"skipkeybdlib" to avoid upgrade of stock keyboard lib (can be used as workaround for recent CyanogenMod keyboard issues)
"forceclean" to enforce clean install

### DIFF
--- a/scripts/inc.compatibility.sh
+++ b/scripts/inc.compatibility.sh
@@ -27,7 +27,8 @@ keyboardlibhack(){ #only on lollipop and higher
 /system/app/LatinIME/lib/$ARCH/libjni_latinimegoogle.so"
     KEYBDLIBS='keybd_lib_filename1="libjni_latinimegoogle.so";
 keybd_lib_filename2="libjni_latinime.so";'
-    KEYBDINSTALLCODE='if ( ! contains "$gapps_list" "keyboardgoogle" ); then
+    # Do not touch AOSP keyboard if it's neither removed or replaced by Google's one
+    KEYBDINSTALLCODE='if ( ! contains "$gapps_list" "keyboardgoogle" ) && ( ! contains "$gapps_removal_list" "keyboardstock" ) && [ "$skipkeybdlib" = "false" ]; then
     extract_app "Optional/keybdlib";
     ln -sf "/system/'"$LIBFOLDER"'/$keybd_lib_filename1" "/system/'"$LIBFOLDER"'/$keybd_lib_filename2"; # create required symlink
     mkdir -p "/system/app/LatinIME/lib/'"$ARCH"'";
--- a/scripts/inc.updatebinary.sh
+++ b/scripts/inc.updatebinary.sh
@@ -89,7 +89,7 @@ contains() {
 }
 
 clean_inst() {
-  if [ -f /data/system/packages.xml ]; then
+  if [ -f /data/system/packages.xml ] && [ "$forceclean" != "true" ]; then
     return 1;
   fi;
   return 0;
@@ -575,6 +575,20 @@ if ( grep -qiE "forcedpi(120|160|213|240|280|320|400|480|560|640|nodpi)" $g_conf
   density=${density#forcedpi};
 fi;
 
+# Check for Clean Override in gapps-config
+if ( grep -qiE "^forceclean$" $g_conf ); then # true or false to override the default selection
+  forceclean="true"
+else
+  forceclean="false"
+fi;
+
+# Check for keybdlib Override in gapps-config
+if ( grep -qiE "^skipkeybdlib$" $g_conf ); then # true or false to override the default selection
+  skipkeybdlib="true"
+else
+  skipkeybdlib="false"
+fi;
+
 # Set density to unknown if it's still empty
 test -z "$density" && density=unknown;
 
@@ -982,13 +996,19 @@ for gapp_name in $core_gapps_list; do
   get_appsize "Core/$gapp_name";
   core_size=$((core_size + appsize));
 done;
-unzip -o "$ZIP" "Optional/keybd_lib.tar.xz" -d /tmp;
-keybd_lib_size=$(tar -tvJf "/tmp/Optional/keybd_lib.tar.xz" "keybd_lib" 2>/dev/null | awk 'BEGIN { app_size=0; } { file_size=$3; app_size=app_size+file_size; } END { printf "%.0f\n", app_size / 1024; }');
-rm -f "/tmp/Optional/keybd_lib.tar.xz";
 
-# Determine final size of Core Apps
+# Add Keyboard Lib size to core, if it exists
 if ( ! contains "$gapps_list" "keyboardgoogle" ); then
+  unzip -o "$ZIP" "Optional/keybd_lib.tar.xz" -d /tmp;
+  keybd_lib_size=$(tar -tvJf "/tmp/Optional/keybd_lib.tar.xz" "keybd_lib" 2>/dev/null | awk 'BEGIN { app_size=0; } { file_size=$3; app_size=app_size+file_size; } END { printf "%.0f\n", app_size / 1024; }');
+  rm -f "/tmp/Optional/keybd_lib.tar.xz";
   core_size=$((core_size + keybd_lib_size)); # Add Keyboard Lib size to core, if it exists
+fi
+
+# Do not touch AOSP keyboard if it's neither removed or replaced by Google's one
+if ( ! contains "$gapps_list" "keyboardgoogle" ) && ( ! contains "$gapps_removal_list" "keyboardstock" ) && [ "$skipkeybdlib" = "true" ]; then
+  reqd_list=$(echo "${reqd_list}" | grep -v "latinime.so");
+  remove_list=$(echo "${remove_list}" | grep -v "latinime.so");
 fi
 
 # Read and save system partition size details


scripts/inc.compatibility.sh
```diff
@@ -27,7 +27,8 @@ keyboardlibhack(){ #only on lollipop and higher
 /system/app/LatinIME/lib/$ARCH/libjni_latinimegoogle.so"
     KEYBDLIBS='keybd_lib_filename1="libjni_latinimegoogle.so";
 keybd_lib_filename2="libjni_latinime.so";'
-    KEYBDINSTALLCODE='if ( ! contains "$gapps_list" "keyboardgoogle" ); then
+    # Skip keyboard treatment on CyanogenMod without Google version (CM's LatinIME is broken with Google lib)
```
mfonville: Are you sure that CM's LatinIME is always broken with the Google lib?
We had some reports of some issues sometimes, but for other people it does seem to work :-/

mfonville:
Thank you for your pull request.
The keyboardlib handling is useful, but I wonder where you did get the information about the broken lib in combination with CM? Because we get very mixed reports from users.

About the custom removals, I think it would be better if that would be handled by the existing mechanism of Universal Removal. See https://github.com/opengapps/opengapps/wiki/Advanced-Features-and-Options#universal-application-removals which is more 'safe', as it only checks for applications in /app and /priv-app

Also, if possible, I would like to avoid introducing a second configfile, because that would complicate things even more, e.g. we already check over 10 different combinations for the current config location, it would realloy make things more complex to determine which configfile is used.

I do like the idea of being able to force a "clean install" behaviour (though explicitly mentioning that this 'unsupported', one should not report bugs with camera/keyboard in that case). We could offer that then through e.g. a "ForceClean" keyword in the regular configfile.

@rapperskull do you have any extra thoughts?

rapperskull:
I also think that another removal feature is redundant, and dangerous too. All that can be done with the current config file should be handled that way, and overwriting installer.data lists is not very wise. The ForceClean feature is indeed good if handled with current gapps-config.
As for as the keyboard changes, we have to first confirm the CM bug.

eugenesan:
Thank you for prompt responses.
After reviewing referenced information, I agree that new "override" config is redundant.
As for the rest:
a. I thing ability to add variables declaration into gapps-config might be very useful and that can be implemented by parsing something like [VAR="value"] in gapps-config.
b. CM's AOSP keyboard started misbehaving only few days ago and from the logcat it seems like new Google lib lacks dictionary support (com.android.inputmethod.dictionarypack.*).

mfonville:
Including variable declaration with "./" execution is IMHO quite dangerous, since it allows for arbitrary/unchecked code to run. So, while powerful, if one would need this kind of functionality, they could better build their own GApps package, than the prebuilt ones from opengapps.org

About the keyboardlib, should I maybe revert the latest keyboardlib that is used? And could you maybe put that logcat in a gist, then I can also study it.

Would you be wiling update the pull request based on these comments?

eugenesan:
There is no code sourcing in 2nd iteration of the PR and I can add syntax validation to avoid any chance of code execution.

Regarding kbdlib, I guess "pinning" lib version is not a good idea.
Anyway, updating kbdlib in regardless of installing keyboard component sounds odd.
Especially when some users report problems with it.
Maybe kbdlib should be separated into standalone component to allow control of it's installation by user.

mfonville:
I did consider at some moments to drop kbdlib completely, since indeed troubleshooting it is hard. But it does add the feature (gestures) that users "expect" of (most) GApps packages. So I prefer not to drop it, if possible. But I did exclude it e.g. already on the 4.4 packages.

Maybe Google Keyboard will have less issues in the future for "dirty" installations, which should lower the pressure to bundle this lib, and e.g. making it an addon.

scripts/inc.updatebinary.sh
```diff
@@ -17,11 +17,8 @@ echo '#!/sbin/ash
 # This Open GApps Shell Script Installer includes code derived from the TK GApps of @TKruzze and @osm0sis,
 # The TK GApps are available under the GPLv3 from http://forum.xda-developers.com/android/software/tk-gapps-t3116347
 #
-unzip -o "$3" '"$EXTRACTFILES"' -d /tmp;'> "$build/META-INF/com/google/android/update-binary"
```
mfonville: We use $EXTRACTFILES on purpose, because the set of files can/will change per architecture and possible new package variants.

scripts/inc.compatibility.sh
```diff
@@ -26,8 +26,10 @@ keyboardlibhack(){ #only on lollipop and higher
 /system/app/LatinIME/lib/$ARCH/libjni_latinime.so
 /system/app/LatinIME/lib/$ARCH/libjni_latinimegoogle.so"
     KEYBDLIBS='keybd_lib_filename1="libjni_latinimegoogle.so";
-keybd_lib_filename2="libjni_latinime.so";'
-    KEYBDINSTALLCODE='if ( ! contains "$gapps_list" "keyboardgoogle" ); then
+keybd_lib_filename2="libjni_latinime.so";
+keybd_lib_upgrade="true";'
+    # Do not touch AOSP keyboard if it's niether removed or replaced by Google's one
```
mfonville: small typo in the comment :) 'neither'

scripts/inc.updatebinary.sh
```diff
@@ -89,7 +86,7 @@ contains() {
 }
 
 clean_inst() {
-  if [ -f /data/system/packages.xml ]; then
+  if [ -f /data/system/packages.xml ] && [ "$enforce_clean" != "true" ]; then
```
mfonville: Maybe rename to "forceclean", then it will coincide with the gapps-config keyword we would like to use.

scripts/inc.updatebinary.sh
eugenesan: This change was not planned.
Probably slipped in during rebase.

scripts/inc.updatebinary.sh
```diff
@@ -414,11 +411,18 @@ if [ "$g_conf" ]; then
   sed -e 's|#.*||g' -e 's/\r//g' -e '/^$/d'  "$g_conf_orig" > $g_conf; # Strip user comments from gapps-config
   user_remove_list=$(awk -F "[()]" '{ for (i=2; i<NF; i+=2) print $i }' $g_conf); # Get users list of apk's to remove from gapps-config
   sed -i s/'([^)]*)'/''/g $g_conf; # Remove all instances of user app removals (stuff between parentheses)
+  user_vars_list=$(awk -F "[][]" '{ for (i=2; i<NF; i+=2) print $i }' $g_conf); # Get users vars from gapps-config
```
mfonville: Could we postpostone this user_vars_list then for now? It is powerful, but people who want to change how the script works, will more probably edit the script directly.

What we could do, is to think about better/more configuration style for the longer term, e.g. with more structure and parsing.

scripts/inc.updatebinary.sh
```diff
@@ -982,15 +986,21 @@ for gapp_name in $core_gapps_list; do
...
-# Determine final size of Core Apps
+# Add Keyboard Lib size to core, if it exists
 if ( ! contains "$gapps_list" "keyboardgoogle" ); then
+  unzip -o "$ZIP" "Optional/keybd_lib.tar.xz" -d /tmp;
+  keybd_lib_size=$(tar -tvJf "/tmp/Optional/keybd_lib.tar.xz" "keybd_lib" 2>/dev/null | awk 'BEGIN { app_size=0; } { file_size=$3; app_size=app_size+file_size; } END { printf "%.0f\n", app_size / 1024; }');
+  rm -f "/tmp/Optional/keybd_lib.tar.xz";
   core_size=$((core_size + keybd_lib_size)); # Add Keyboard Lib size to core, if it exists
 fi
 
+# Do not touch AOSP keyboard if it's niether removed or replaced by Google's one
```
mfonville: typo again

scripts/inc.updatebinary.sh
```diff
@@ -982,15 +986,21 @@ for gapp_name in $core_gapps_list; do
...
+if ( ! contains "$gapps_list" "keyboardgoogle" ) && ( ! contains "$gapps_removal_list" "keyboardstock" ) && [ "$keybd_lib_upgrade" != "true" ]; then
```
mfonville: Are we sure this also goes well then for users "upgrading" their GApps package to the most recent version? They did receive the lib previously from our install, and it should not accidentally be removed by the gapps-remove list.

mfonville:
Oh, maybe I should add. Please don't be discouraged by my comments. I really do like it that you are contributing to the codebase and thinking of possible new features. :-)

scripts/inc.updatebinary.sh
eugenesan: Well the default action is the same as before so it should be ok.

eugenesan:
I see your concerns.
Anyway, if you decide to postpone variables support, whole PR should be postponed since it useless without it.

mfonville:
You can just create a "ForceClean" keyword check/interpret to the script e.g. how ForceDPIXXX is currently handled. Then at least that part can be pulled.

mfonville:
For some reason GitHub didn't give me a notification of that you did do the changes :-/
Anyhow, it looks good so I will merge it. Could you maybe update the wiki to describe your newly added features ( https://github.com/opengapps/opengapps/wiki/Advanced-Features-and-Options ) ?

eugenesan:
Updated Wiki.

mfonville:
Thank you!

mfonville:
FYI as follow-up, we get more reports about the swype issues. So we will drop those lib installations to prevent any issues. We will try to allow Google Keyboard installations also on dirty installs, and hope that it has improved enough that it doesn't give any trouble on dirty installs anymore.

PR #120, opened 2015-09-28
Selective download, tools checking and parallelism

eugenesan:
This is "combo" PR with following changes:
* Allow downloading single architecture using shallow clone
* Export tools checking into it's own include
* Add support for parallelism and settle source code permissions

2nd iteration includes:
* Fetching archs now fetch arm for fallback on x86 archs and shallow clone is now a specific option 
* Tools checking fixed cosmetically
* Parallelism now includes checks and notifications for memory (over)usage.


eugenesan:
Regarding per-ARCH download: Maybe we can at least split ii into 2 groups?
Regarding shallow checkout: I suggest we use shallow be default and add --full option to allow full checkout?

mfonville:
I'd like to keep full history the default, since most people using the command are our packagers, and for them a full APK archive is useful :-)
We can add the shallow option to the README.md though, and suggest it to the newbies who are just looking for a quick checkout.

About the per-arch, we could indeed split it into two. We could make an 'arm' set that downloads arm+all, and a 'full' set (which would be the default).

mfonville:
All merged! Thanks for all your coding effort!
Do you have btw an XDA handle @eugenesan? then I can add you to the OP (if you want).

mfonville:
We should still update the wiki and/or readme about the new download_sources options.

mfonville:
Somtimes I do get btw: CPU[4]/[3] as output. Which is weird, any reason why that could be?

eugenesan:
@mfonville Sure you can contact me at XDA::@eugenesan
Regarding download_sources options, I couldn't find related wiki content so I'll create PR with updated readme.
And regarding CPU prints it's expected behavior but I agree it's ugly. It's actually prints number of threads including the one that is postponed.

eugenesan:
@mfonville Seems like there is a "by1" bug in CPU reporting, PR is on it's way.

PR #126, opened 2015-10-02
Add failover to non-parallel build mode

eugenesan:


scripts/inc.packagetarget.sh
```diff
@@ -85,6 +85,21 @@ createzip() {
...
+  MEMORY_MIN=800000 # Minimum of RAM required (for single thread) on x86_64 machine [~701MB for xz, 2*25KB for bash and some spare]
+  THREADS="$(($(cat /proc/cpuinfo | grep "^processor" | wc -l)))"
+
+  if ! grep -q "MemAvailable:" /proc/meminfo; then
+    MEMORY=0
+  else
+    MEMORY="$(($(grep "MemAvailable:" /proc/meminfo | awk '{print $2}') / 4))"
+  fi
+
+  if [ $MEMORY = 0 ] || [ $MEMORY -lt $MEMORY_MIN ]; then
```
mfonville: Just for coding style, in general it is safer to put "" around any values that you compare or assign. So even where it is not per se necessary, we often still do this to prevent possible errors and to keep ourselves trained to do so ;-)

scripts/inc.packagetarget.sh
```diff
@@ -85,6 +85,21 @@ createzip() {
...
+    echo "Your system doesn't have free memory or enable to report it reliably, switching off memory checking and parallel mode."
```
mfonville: "WARNING: Can't establish if enough free memory is available: parallel compression mode disabled."

scripts/inc.packagetarget.sh
```diff
@@ -94,35 +109,39 @@ createzip() {
...
+      # Use parallel mode only if we have memory metric and have more then 1 CPU
+      if [ $THREADS -gt 1 ] && [ $MEMORY -gt 0 ]; then
+        # Wait if we reached RAM or THREADS limit
+        tries=0; while true; do
+          # Count still running createxz instances
+          threads=0; for p in $pidlist; do test -d /proc/$p && threads=$((threads+1)); done
+          memory=$(grep "MemAvailable:" /proc/meminfo | awk '{print $2}')
+
+          # Check if reached our limits (3/4 of RAM or THREADS), wait if we are
+          echo "Resource report: FreeRAM[$memory], InitialFreeRAM[$MEMORY], MinimumFreeRAM[$MEMORY_MIN], Threads[$threads], CPUs[$THREADS], XZ RAM[`for p in $(pgrep xz); do echo -n "$(awk '/VmSize/ { print $2 }' /proc/$p/status) "; done`]"
```
mfonville: Can we 'merge' this output somehow with the XZ-compressor function? At the moment a lot of output over different lines is created, where I think it should be able to fit on a single line per XZ compression thread.

scripts/inc.packagetarget.sh
eugenesan: Generally I agree with you, but some shells refuse to compare correctly quoted scalars.

scripts/inc.packagetarget.sh
eugenesan: No problem.

scripts/inc.packagetarget.sh
eugenesan: Actually this output is more for "wait" reporting.
Anyway, "un-mangling" lines order will require whole UI reimplementation in something like CURSES.

scripts/inc.packagetarget.sh
mfonville: Ok, I didn't know that. But then I only use dash, bash and busybox ash :-)
If you ever spot or encounter an error, you'll know what to do ;-)

scripts/inc.packagetarget.sh
mfonville: I understand that better UI would need e.g. curses.
But if instead of reporting the wait, we could report the start of a new thread, and e.g. passing an optional string parameter to the createxz function, that than print the details within the current output.

scripts/inc.packagetarget.sh
eugenesan: Actually I think I encountered this in one of those 3 :-)
I guess older bash and/or busbox's ash built with some option.

scripts/inc.packagetarget.sh
eugenesan: No problem.

mfonville:
Merged, I did some more changes to the output though, to really limit it (to keep also my buildbot logs useful).

PR #130, opened 2015-10-05
Fix installation hangs when flashing with FlashFire.

Asvel:
Thanks for this great project.
This PR fix installation hangs at [Line 573](https://github.com/opengapps/opengapps/blob/b320e48704e344f5f74f3cfeda6943807b158037/scripts/inc.updatebinary.sh#L573) when flashing with FlashFire and gapps-config not exists.
I have noticed only TWRP is supported in the FAQ, but there are several devices that custom recovery isn't available, unless unlock the bootloader (and unlocking may cause some feature stop working, such as Galaxy S6), so I hope an on-device flasher can also be basically supported.

### DIFF
--- a/scripts/inc.updatebinary.sh
+++ b/scripts/inc.updatebinary.sh
@@ -417,6 +417,8 @@ if [ "$g_conf" ]; then
   sed -i '/^$/d' $g_conf; # Remove all empty lines for cleaner appearance
 else
   config_file="Not Used";
+  g_conf=/tmp/proc_gconf;
+  touch "$g_conf";
 fi;
 
 # Unless this is a NoDebug install - create folder and take 'Before' snapshots


mfonville:
Ofc we will try to stay compatible with other recoveries, if possible.
Could you explain to me what triggers that the installation hangs if no g_conf file is found? Just so that I understand what the technical reason is for the line, before I merge it :-)

Asvel:
Thanks for your efforts.
The problem line was located by several `ui_print`, I think the reason maybe when `$g_conf` is empty string, the grep call become `grep -qiE "forcedpi(120|160|213|240|280|320|400|480|560|640|nodpi)"`, then it waiting input from stdin forever.
So I make a quick fix by creating a fake gapps-config file. But for now, I think the better solution could be surrounding all `$g_conf` with `""`.

mfonville:
Ok, I understand. I will merge your "prevention" fix then, and I'll try to clean up the unquoted $g_conf references (and a note to @opengapps/developers that we should not forget about those ```""``` :) )

Asvel:
Thanks :)

mfonville:
@Asvel no, thank you! We need people like you, that test, research bugs, fix them and make PRs :-)

PR #133, opened 2015-10-06
Use 'nproc' to get the number of available processors

mc1100:


mfonville:
Thanks @mc1100 for the PR and @rapperskull for reviewing :)

Issue #132, opened 2015-10-06
Creation of a "super" package


mfonville:
Since Google did drop some apps from the Nexus stock images, but we do have official names for them, and people installing and depending on them e.g. Google Earth, we can make a "super" package that contains these apps, and keep "stock" equal the latest offical Google stock experience.
We can also include the less used packages, like international keyboards, Google Fi, Android for Work etc in this package. And less/in-compatible apps, that people might want to experiment with, like Android Pay (see #116) and maybe even Google Dialer.
I did discuss this idea with @rapperskull and it seems at the moment more easy to implement such a big package than creating seperate add-ons, which would require much more changes to our code, build and website infrastructure.

- [x] Collect APKs of all apps we want to bundle in super
- [x] Update path and lib definitions for these APKs in inc.buildtarget.sh
- [x] Let gapps-remove.txt be generated based on this "super" pakage instead of "stock"
- [x] Update package set definitions in inc.buildtarget.sh
- [x] Update the website to allow downloading of this "super" package
- [x] Update the wiki for contents op both "super" and "stock"
- [x] Post on XDA what has changed and update OP and and the hotlinks
- [x] Possibility of a better name than "super" (any suggestions can be posted in the comments)

rapperskull:
I just made a list of apps contained in the Nexus 6 factory image and filtered only the ones with the package name containing "google". We should now filter only the ones we don't already include and select the ones worth adding.

File path | Package name
--------- | ------------
/system/app/Books/Books.apk | com.google.android.apps.books
/system/app/CalculatorGoogle/CalculatorGoogle.apk | com.google.android.calculator
/system/app/CalendarGooglePrebuilt/CalendarGooglePrebuilt.apk | com.google.android.calendar
/system/app/CloudPrint2/CloudPrint2.apk | com.google.android.apps.cloudprint
/system/app/DMAgent/DMAgent.apk | com.google.android.apps.enterprise.dmagent
/system/app/Drive/Drive.apk | com.google.android.apps.docs
/system/app/EditorsDocs/EditorsDocs.apk | com.google.android.apps.docs.editors.docs
/system/app/EditorsSheets/EditorsSheets.apk | com.google.android.apps.docs.editors.sheets
/system/app/EditorsSlides/EditorsSlides.apk | com.google.android.apps.docs.editors.slides
/system/app/FitnessPrebuilt/FitnessPrebuilt.apk | com.google.android.apps.fitness
[...]

PR #145, opened 2015-10-17
[AROMA] Update for Super and Various Cleanup

bgiesing:
For issue #132 

- Add new Super apps
- Make title in installer match the Play Store name
- Avoid duplicate text from description and title.
- Add Swype Libs optional feature
- Change XDA link to homepage as it's home to that and more.

@mfonville @raulx222 Do I have it right? Don't have a device to test but it should work.


mfonville:
Thanks! Also really nice that you improved all those other texts/descripitions, order and the swypelibs keyword!

bgiesing:
@mfonville Fixed!

PR #146, opened 2015-10-18
[AROMA] Fix PlayGames and ProjectFi Install

bgiesing:
@mfonville Accidentally messed up the Play Games and Project Fi keywords so tonight's build won't install them (without rebuild)!


mfonville:
I will rebuild them!

PR #152, opened 2015-10-24
tools: Fix Zipalign Check

javelinanddart:
* The Version of Zipalign used to build Lollipop doesn't have this
  When I removed this not check it still build just fine
  Thus, Remove It

### DIFF
--- a/scripts/inc.tools.sh
+++ b/scripts/inc.tools.sh
@@ -33,7 +33,7 @@ checktools() {
     else
       case $command in
         zipalign)
-          if ! zipalign 2>&1 | grep -q "page align stored shared object files"; then
+          if zipalign 2>&1 | grep -q "page align stored shared object files"; then
             echo "zipalign is outdated. Install a recent version from the Android SDK." >&2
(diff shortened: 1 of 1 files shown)


mfonville:
It will build, but your Chrome will be completely broken, it needs its 'crazy' libs inside and aligned.
Next to that it will be a requirement for any app when building Marshmallow, because then all libs will have to be inlined and aligned.

Lib-alignment was added for a reason to the newer zipaligns.

So sorry, I won't be pulling this one.

javelinanddart:
Chrome works just fine #1,
#2, where can i get the required zipalign so i don't have to remove this?

mfonville:
Then your currently active Chrome is probably read from /data/ because if the one installed on /system/'s libs are not aligned, it cannot start. That is why it was added in the first place (https://github.com/opengapps/opengapps/commit/0142095ef8ecf53656d192deaea13df6be12868c before that commit, for a day all Chrome installations were broken).

You can install it (as literally suggested by the script) from the Android SDK.

PR #161, opened 2015-11-08
installdata: remove PackageInstaller if its in priv-app too

invisiblek:


invisiblek:
it might be better to leave app/ and add priv-app/, just to make sure it gets killed off

mfonville:
Yeah. Both locations is safer. If you'd update the patch I will pull that

mfonville:
Thanks!

airend:
Thanks, @invisiblek! I thought I was going crazy with things not finishing booting, but all is finally well with your fix.

PR #181, opened 2015-12-05
Add Chromium to browser removal list

Great-Geek:
Removes prebuilt Chromium which ships with custom ROMs that decided the AOSP Browser was bad


mfonville:
Thanks! Good improvement :-)

PR #187, opened 2015-12-09
installdata: remove runtime-permissions.xml

invisiblek:
* This xml holds all the permissions for system apps and is only
  generated once (aka never updated). If a user flashes and boots
  a rom without gapps installed, this xml will get generated without
  the permissions set for gapps (makes sense, since they aren't
  there) so adding gapps later will cause force closes all over
  since gapps wont have their permissions.

* Killing this xml at gapps install time will ensure it gets
  re-generated during the next boot and gapps will have their proper
  permissions.

### DIFF
--- a/scripts/inc.installdata.sh
+++ b/scripts/inc.installdata.sh
@@ -411,7 +411,9 @@ app/WhisperPush'"$REMOVALSUFFIX"'
 # _____________________________________________________________________________________________________________________
 #                                             Permanently Removed Folders
 # Pieces that may be left over from AIO ROMs that can/will interfere with these GApps
+# Also remove the runtime-permissions.xml in case the rom had been booted prior to installing GApps
 other_list="
+/data/system/users/*/runtime-permissions.xml'"$REMOVALSUFFIX"'
 /system/app/CalendarGoogle'"$REMOVALSUFFIX"'
 /system/app/CloudPrint'"$REMOVALSUFFIX"'
 /system/app/DeskClockGoogle'"$REMOVALSUFFIX"'


invisiblek:
This could probably go through a good barrage of testing before shipping, but it appears to be working as expected here.

mfonville:
Thank you for your PR.

So if I understand correctly, this file is not changed by any interactions of the user with the permissions system? i.e. when changing any of the permission settings of one of the (G)apps?

Because if a users removes purposefully some permissions of an app, we shouldn't revert those.

About the contents of the commit:
'"$REMOVALSUFFIX"' is not necessary in this line, it contains ".apk" on Android 4.4 and is empty for higher versions, and is there to provide compatibility with KitKat where apps are not stored in subfolders.

In general we could make this line somewhere more explicit instead of grouping it with $other_list.

Try not to include the "sources/****" paths in the commit. (in .gitmodules it is also specified that those should be ignored, we only commit those when we e.g. change the structure of the sources-directories).

@rapperskull could you also have look at this and share your thoughts?

rapperskull:
Unfortunately that file is changed every time you grant/revoke a permission to an app, so I don't think it's a feasible solution.

invisiblek:
Hmm, was afraid of that. I'll look in to some other options. At the very least it gets the conversation going in the right direction to get this fixed.

mfonville:
Yeah, +1 to look at options to fix this.

Maybe we could grep or manipulate the file in some way?

invisiblek:
Grep wouldn't work. We'd have to actually add to this file. Via shell scripting would be painful (but probably doable) but we'd need to somehow generate these permissions dynamically based on which size package was installed.

mfonville:
Would an approach like this maybe work?
* Get all existing permissions from the file
* Filter out all permissions that are GApps-related, keep the non-GApps seperate for a moment
* Keep all current existing GApps permissions, (re)add all GApps permissions that are not set to the file
* Glue the non-Gapps back to the file

invisiblek:
That seems pretty reasonable

mfonville:
@invisiblek 
Could you try if https://github.com/opengapps/opengapps/tree/mmpermissions works for you?

mfonville:
I am re-implementing this trick in a new manner; Only doing this runtime permissions removal on ROMs that are dirty flashing GApps for the very first time

PR #219, opened 2016-01-10
Fix skipswypelibs and add related UI prints and logging

eugenesan:
- Fix skipswypelibs by fixing minor logic bug in f429c999999509236aa6dcdd1ef2263d0fbeaa33
- Add related UI prints and logging
- Improve existing UI prints and logging in lines surrounding affected code


mfonville:
Thanks!

PR #223, opened 2016-01-13
PackageInstaller: move into PicoGApps, making it removable

KreAch3R:
Last relevant commits moved PackageInstaller into core GApps, making it impossible to exclude with a .gapps-config keyword or AROMA.
In the meantime, the open gapps GooglePackageInstaller is buggy: http://i.imgur.com/lhAZnBd.png (Stock nexus rom's PackageInstaller doesn't do that) (DPI bug?)

This commit:

- Moves PackageInstallerGoogle from core to Stock package
- Makes it configurable with tag "PackageInstallerGoogle"
- Adds it as an option in the AROMA version (untested by me, followed previous commits' structure)
- RESULT: http://i.imgur.com/HEcUlWY.png with a Open-GApps STOCK package

### DIFF
--- a/scripts/inc.compatibility.sh
+++ b/scripts/inc.compatibility.sh
@@ -263,10 +263,9 @@ webviewstock"
 
 api23hack(){
   if [ "$API" -ge "23" ]; then
-    gappscore="$gappscore
-packageinstallergoogle"
     gappspico="$gappspico
-googletts"
+googletts
+packageinstallergoogle"
     gappsmini="$gappsmini
 calculatorgoogle"
     gappsstock="$gappsstock
--- a/scripts/inc.aromadata.sh
+++ b/scripts/inc.aromadata.sh
@@ -170,6 +170,7 @@ form(
       "Music",     "<b>Google Play Music</b>",       "",                      "check",
       "NewsStand",     "<b>Google Play Newsstand</b>",       "",                      "check",
       "NewsWidget",     "<b>Google News & Weather</b>",       "",                      "check",
+      "PackageInstallerGoogle",     "<b>Google PackageInstaller</b>",       "",                      "check",
       "Pinyin",     "<b>Google Pinyin Input</b>",       "",                      "check",
       "Photos",     "<b>Google Photos</b>",       "",                      "check",
       "PlayGames",     "<b>Google Play Games</b>",       "",                      "check",
@@ -511,6 +512,12 @@ if
   prop("gapps.prop", "NewsWidget")=="1"
 then
   appendvar("gapps", "NewsWidget\n");
+endif;
+
+if
+  prop("gapps.prop", "PackageInstallerGoogle")=="1"
+then
+  appendvar("gapps", "PackageInstallerGoogle\n");
 endif;
 
 if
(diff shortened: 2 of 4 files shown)


mfonville:
Thanks for the PR.
I will give some feedback on the commit, to hopefully improve the PR code-wise.
Meanwhile I would like @rapperskull for some technical feedback. I.e. do we know of any adverse effects of people are *not* using Google's PackageInstaller?

scripts/inc.compatibility.sh
```diff
@@ -263,15 +263,13 @@ webviewstock"
...
   if [ "$API" -ge "23" ]; then
-    gappscore="$gappscore
-packageinstallergoogle"
     gappspico="$gappspico
 googletts"
     gappsmini="$gappsmini
 calculatorgoogle"
     gappsstock="$gappsstock
-contactsgoogle"
-#dialergoogle"
```
mfonville: Can you keep the #dialergoogle line there?
The idea is that it could be easily activated in the future when we will have the issues with the dialer sorted out.

scripts/inc.compatibility.sh
```diff
@@ -263,15 +263,13 @@ webviewstock"
 
 api23hack(){
   if [ "$API" -ge "23" ]; then
-    gappscore="$gappscore
-packageinstallergoogle"
     gappspico="$gappspico
 googletts"
     gappsmini="$gappsmini
 calculatorgoogle"
     gappsstock="$gappsstock
```
mfonville: I think it might be in general a good idea to treat PackageInstaller as we do with calsync, since I consider it quite a "core" app. Would you find it a good idea to add it to "gappspico" instead? It will then still be keyword configurable, but included for 99% of the users.

sources/all
```diff
@@ -1 +1 @@
-Subproject commit 47f59d8b7c5cfd3690f1e7fdba0c374e7b45dbd4
+Subproject commit 66c1cb4f5e2f6d7e4deb94177ad915407cc07442
```
mfonville: Please try to exclude any of the sources/* entries in the commit

scripts/inc.compatibility.sh
KreAch3R: Yes, no problem :)

scripts/inc.compatibility.sh
KreAch3R: That's perfect for me, I just want it to be keyword configurable. Should I add it after googletts in L270 above?

sources/all
KreAch3R: Oops, sorry for that. The new commit won't have them

KreAch3R:
Thank you for the fast reply @mfonville ! I'll rework the commit with your additions. Regarding your question, I do not know of any adverse effects. I'm talking as a ROM dev, a lot of my users use other GApps packages like BaNks which don't include it and they have no problems with their gapps usage.

scripts/inc.compatibility.sh
mfonville: :+1:

KreAch3R:
Just updated the commit :)

rapperskull:
Just looked at this PR, but I can't understand what's the exact problem. The orange bar with "NEW ALL"?
IMHO if we're sure that the AOSP package installer works fine too, we can merge this PR without problems. One of the key points of OpenGApps is the customization, so if the user wants to exclude GooglePackageInstaller, he should be able to, if it's not essential.

KreAch3R:
:+1: 
 
@rapperskull It was both what you said about open-gapps being about customization and the fact that the current GooglePackageInstaller is bad. I love the options Open-GApps give you and I am actively promoting them with my ROM but I had a lot of users 1) blaiming the ROM for Google's PackageInstaller, because they didn't know it was installed 2) hated the fact that they have to live with that orange bar in PackageInstaller, without being able to uninstall it. 

So they were switching to other gapps packages. 

Regarding AOSP PackageInstaller and problems. It must surely work OK because Open-GApps is the only one that bundles it in, AFAIK. Besides that, I am running my ROM these last five days with the stock package I created (w/out GooglePackageInstaller) and I have no problems at all.

So, this merge is great! Thank you! :beers:

mfonville:
:-)

PR #232, opened 2016-01-23
Enable GoogleDialer(Phone)

eugenesan:
New GoogleDialer is Java only app which confirmed to work even on old HTC-Sense (Pyramid).
If user will report issues we could make it disabled by default.


mfonville:
There were many issues reported with it not working. Often at the moment that a call was actually made (or the phone ringing). Did you test those situations with pyramid?

In general I am thinking we should start with a whitelist mechanism.

eugenesan:
@mfonville You were right, the dialer didn't stand the real call test and persistently caused soft-reboot.

PR #233, opened 2016-01-23
Add new CM's Snap camera to stock cameras list

eugenesan:


scripts/inc.installdata.sh
```diff
@@ -221,6 +221,7 @@ priv-app/Camera'"$REMOVALSUFFIX"'
 priv-app/Camera2'"$REMOVALSUFFIX"'
 app/MotCamera'"$REMOVALSUFFIX"'
 priv-app/MotCamera'"$REMOVALSUFFIX"'
+app/Snap/Snap'"$REMOVALSUFFIX"'
```
mfonville: are you sure it is twice in a subdirectory?
shouldn't it be just app/Snap and not app/Snap/Snap?

eugenesan:
Fixed redundant subfolder, thanks for pointing out.

mfonville:
:-)

Lgtm now

gericom-hummer:
Is it possible to leave us a choice to bypass stock camera when installing Google cam?

PR #234, opened 2016-01-23
Support maximal API per package and limit GoogleCamera to 19

eugenesan:
Recent GoogleCamera 3.x will fail on every device released before Marshmallow due to missing CameraAPI2. Nexus users and owners of newest phones will be automatically upgraded to version 3.x using GooglePlay.

### DIFF
--- a/scripts/inc.buildhelper.sh
+++ b/scripts/inc.buildhelper.sh
@@ -118,8 +118,9 @@ buildapp() {
   else usearch="$4"; fi #allows for an override
 
   minapihack #Some packages need a minimal api level to maintain compatibility with the OS
+  maxapihack #Some packages need a maximal api level to maintain compatibility with the OS
 
-  if getapksforapi "$package" "$usearch" "$API" "$useminapi"; then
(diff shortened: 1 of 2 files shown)


mfonville:
Thanks for this PR. I am still ill and stuck in my bed (and hence no real coding possible).
I had something like this in my mind too. I was thinking to use the same 'trick' to add e.g. the v2 camera as CameraGoogleLegacy as extra line in buildtarget for api23. And then let the installerscript automatically use it as fallback on non-v3-compatible devices.

eugenesan:
I see.
Some kind of new trick is required here. It's a bit sad to add one when we already have plenty...
Well, I'll have to think about it.

mfonville:
I did not find a reliable way yet to check if the camera3 HAL is available/compiled (https://source.android.com/devices/halref/camera3_8h_source.html) :-(

I checked both for the existance of certain libs on /system/ as their library properties using readelf -d.

mfonville:
Track this branch instead for progress: https://github.com/opengapps/opengapps/tree/legacycamera

eugenesan:
:+1:

PR #238, opened 2016-01-25
update bundledawk branch with recent changes on master

championswimmer:


championswimmer:
@mfonville please merge this. will make bundledawk branch uptodate with master

mfonville:
There werd still some possible issues my bundledawk branch, because toybox also possible also failed some other commands.
I need to do some more testing before I can merge this (and I need to e.g. update the license in the Buildbot)

championswimmer:
Maybe a prebuilt busybox in the recovery would be a one-stop  solution. 

Will have to change the 'awk' commands to 'busybox awk' in the scripts too.

mfonville:
That is indeed the current path on investigation. Osmosis had shared an interesting busybox commands override mechanism that I want to try :-)

mfonville:
We will go the busybox way; so I will close this PR.

PR #263, opened 2016-02-15
Include our own Busybox fixes #211

### LINKED ISSUE #211: Use bundled busybox as Toybox replacement
toybox doesn't offer awk, but opengapps installer uses it.

mfonville:
To prevent anymore issues with broken environments, we include our own busybox.
This fixes issue #211
Sometimes CM recoveries are still crappy (as on my i9300) but it will fail gracefully in those situations.


mfonville:
I will update the FAQ accordingly for the new ERROR 10 definition.

PR #286, opened 2016-03-27
Correct opengapps.org params for architectures.

jkdf2:
@opengapps/developers


mfonville:
Thanks for spotting that!

PR #290, opened 2016-04-01
Create de.lang

xdevs23:
Fixes Nothing.

Changes:
- Add german language to aroma res
-
-

@opengapps/developers

### DIFF
--- a/scripts/aroma-resources/langs/de.lang
+++ b/scripts/aroma-resources/langs/de.lang
@@ -1,0 +1,57 @@
+### LICENSE:
+#
+# Insert license here :D
+#
+### LANGUAGE STRING RESOURCE FILE
+#
+#   Language  : German
+#   By        : xdevs23
+#
+
+#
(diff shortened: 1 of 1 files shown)


mfonville:
@raulpetru could you take a look at this?

mfonville:
We were not aware that AROMA-installer already contained several language files. We will merge them ASAP.

xdevs23:
I think these are not the full translations

mfonville:
@xdevs23 They are indeed not, they only cover the aspects of AROMA-Installer itself, not our aroma-specific config.

Unfortunately Aroma-Installer is no longer maintained. We are already waiting ages on a new one based on the new libaroma; but @raulpetru is also planning a web-based config-generator, that should make things more easy and translatable.

xdevs23:
Alright. If I do not notice once it is fully translatable, please notify me so I can translate it :)

mfonville:
will do :+1:

xdevs23:
Great.

PR #292, opened 2016-04-03
remove YuBrowser in Mokee ROM

Acris:
Fixes nothing.

Changes:
- remove YuBrowser in Mokee ROM

@opengapps/developers


mfonville:
thanks!

PR #295, opened 2016-04-04
Fix

AlexLartsev19:
Fix mms in "Optional Stock/AOSP/ROM Removal List".

Changes:
- Nothing

@opengapps/developers

Add mms removial list to "Optional Stock/AOSP/ROM Removal List"


mfonville:
Thanks!

PR #304, opened 2016-04-21
Removed redundant file path pittpatt

robin-vandaele:
Changes:
- Removed redundant pittpatt

@opengapps/developers

My knowledge of scripts isn't that big, but I guess this was redundant?


mfonville:
yes, thanks for spotting that!

PR #309, opened 2016-05-03
Fallback to script wrappers if symlinking failed when installing busybox

### LINKED ISSUE #299: Cyanogenmod Recovery support (selinux)
**Checklist**

- [x] The issue is not answered by reading the [FAQ](https://github.com/opengapps/opengapps/wiki/FAQ)
- [x] The issue is part of The Open GApps Project scope *(and not of the pre-built OpenGApps.org packages scope, for support on those see [XDA Forum](http://forum.xda-developers.com/android/software/Open-GApps-t3098071))*

*If Open GApps installation fails:*
- [x] Is the issue reproducible using the latest version of your recovery?
- [ ] Is the issue reproducible with (the latest) [TWRP](https://twrp.me) recovery? *(other recoveries are only best-effort)*

**Issue description:**
As I found out in https://github.com/opengapps/opengapps/issues/278#issuecomment-205008056 , the reason why Open GApps is failing installing on Cyanogenmod Recovery is because of selinux.

I only opened this issue, so we don't forget it.

Kristian

gentoo-root:
Fixes #299.

@opengapps/developers @mfonville

Some recoveries (e.g. CyanogenMod recovery) run strict SELinux policies
that disallow creating symlinks and hardlinks in /tmp and setting
permissive mode. In such case install BusyBox applets as shell script
wrappers instead of symlinking them.

### DIFF
--- a/scripts/inc.installer.sh
+++ b/scripts/inc.installer.sh
@@ -94,11 +94,14 @@ if [ -e "$bb" ]; then
   install -d "$l"
   for i in $($bb --list); do
     if ! ln -sf "$bb" "$l/$i" && ! $bb ln -sf "$bb" "$l/$i" && ! $bb ln -f "$bb" "$l/$i" ; then
-      echo "ui_print ERROR 10: Failed to set-up Open GApps'"'"' pre-bundled '"$2"'" > "$OUTFD"
-      echo "ui_print" > "$OUTFD"
-      echo "ui_print Please use TWRP as recovery instead" > "$OUTFD"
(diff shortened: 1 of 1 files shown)


mfonville:
@gentoo-root did you test if it works correctly?

gentoo-root:
@mfonville yes, I've just installed opengapps successfully on my phone with CM recovery and this patch. Works fine, but google services crashing (seems to be irrelevant to this bug).

mfonville:
@gentoo-root :+1: great, thanks for implementing this!

PR #315, opened 2016-05-10
Whitelist OnePlus One (bacon) for Google Camera

saschpe:
@opengapps/developers This one is slightly more controversial. I am able to do photo spheres, hdr+, panorama and lens blur on CM13 nightly. However, video recording reliably crashes the app. 

For the first 10 - 15 minutes of using the app, I got occasional crashes too. Strangely, after a while I got no crashes and now it works reliably (minus video recording). Maybe related to outdated data elsewhere, dunno.

So the question is if that's enough to not having people to force things.

### DIFF
--- a/scripts/inc.installer.sh
+++ b/scripts/inc.installer.sh
@@ -1303,7 +1303,7 @@ else
   else
     # If not explictly defined, check whitelist
     case $device_name in
-      ryu|angler|bullhead|shamu|volantis*|flounder*|hammerhead*|sprout*) newcamera_compat="true[whitelist]";;
+      ryu|angler|bacon|bullhead|shamu|volantis*|flounder*|hammerhead*|sprout*) newcamera_compat="true[whitelist]";;
       *) newcamera_compat="false";;
     esac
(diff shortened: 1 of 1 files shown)


mfonville:
@saschpe that video doesn't work reliably is caused because the drivers don't support the new API. I'd prefer not to merge this until updated drivers are released and used in the ROMs for this device.

saschpe:
Fine with me. I only found https://jira.cyanogenmod.org/browse/BACON-4283, there where other issues marked as fixed but it seems some people don't have the issue https://forum.cyanogenmod.org/topic/119193-video-mode-on-google-camera-app-cant-connect-to-camera/.

mfonville:
Recreation is probably not possible for users of small GApps packages, because Play Store won't offer the new camera to 'incompatible' devices X-D.

In am not aware of any devices besides those from Google themselves that support the new API yet. And I doubt many older devices ever will, because the camera drivers that often come directly from some OEM vendors need new features implemented to satisfy it. :/

PR #316, opened 2016-05-11
Added A0001 variant for Oneplus One

robin-vandaele:
Fixes dialer not installing an A0001 variant of Oneplus One (We still have to set it default in the settings right?).

Changes:
- Adds A0001 to the whitelist

@opengapps/developers

The A0001 variant also is copatible with GoogleDialer


mfonville:
FYI http://opengapps.org/blog/post/2016/05/13/google-dialer-for-all/

PR #318, opened 2016-05-20
ShellCheck fixes

Bracket-:
NEEDS TESTING AND REVIEW

This PR is completely untested and blind, though the changes should be relatively safe.

Changes:
- Added ShellCheck directives to find sourced files
- Blind fixes to silence a few ShellCheck warnings
  - 85ce8a9 assumes you meant to treat the variables as strings (looks right since sed and cut are used)
  - 7cf01a5 $f doesn't exist AFAIK. Appears $file was meant to be used

I used shellcheck source=scripts/* in scripts/build_gapps.sh so recursively running ShellCheck from the projects root directory works. ShellCheck doesn't appear to offer a nice way of handling this. Let me know if you want it changed.

I'll post a new ShellCheck log in issue #48 for developers that are familiar with the code.

@opengapps/developers

### DIFF
--- a/add_sourceapp.sh
+++ b/add_sourceapp.sh
@@ -39,7 +39,7 @@ installapk() {
 
   if [ "$sdkversion" -lt "$lowestapi" ]; then
     for s in $(seq "$(($sdkversion + 1))" "$lowestapi"); do
-      for d in $(printf "$dpis" | sed 's/-/ /g'); do
+      for d in $(printf "%s" "$dpis" | sed 's/-/ /g'); do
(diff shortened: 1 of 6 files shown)


mfonville:
Thanks for your PR! I will test it and verify it doesn't break anything ;-)

I hope that when we switching to bash we can drop of the some older posix-compatibility hacks and use arrays to fix some of the shellcheck warnings. Still working on that though (it is part of the buildsystem revamp)

mfonville:
ok, I can confirm that it works. Just needed to compile my own shellcheck, because the default shellcheck in Debian/Ubuntu is seriously outdated and doesn't support those annotations yet :(

PR #335, opened 2016-06-26
Update Busybox to 1.25.0

yashdsaraf:


mfonville:
Wow, they are almost half the size of the previous builds?

yashdsaraf:
Oh fu**, I'm so sorry, you use the selinux builds right. Damn it I uploaded the normal ones (;-_-). 
If you want me to, I'll send the correct PR right away!

PR #336, opened 2016-06-26
Update Busybox to 1.25.0 (Corrected)

yashdsaraf:


mfonville:
Thanks for fixing it :)

mfonville:
They are btw still smaller than the older binaries, did you add some extra optimizations? :)

yashdsaraf:
I didn't apply any optimzations, the applet count actually *increased* compared to the last release! :)

PR #365, opened 2016-10-24
Remove StorageManager when installing StorageManagerGoogle

rapperskull:
Changes:
- Remove StorageManager when installing StorageManagerGoogle

I'm not sure that's the right way to do it, so I created a pull request.

@opengapps/developers


mfonville:
handled as separate commits

PR #383, opened 2016-11-13
Don't recursively change perms in /system

luca020400:
Fixes #nothing
Just some sanity in the perms fix

Changes:
- Don't change perms in /system recursively
 -> Default perm is 644


hayateer:
ผมมึนครับ

mfonville:
Thanks for the PR; it needed a bit more work so I have written a new commit with the same idea in mind and leveraging the logic from your other PR: https://github.com/opengapps/opengapps/commit/04be5bf893921e5e3e5cae0886923f6044a77df0

PR #402, opened 2016-12-19
Use APKMirror's MD5 pre-check for duplicates

NicholasBuse:
passes shell check!

### DIFF
--- a/upload_sources.sh
+++ b/upload_sources.sh
@@ -186,7 +186,7 @@ if [ -n "$newapks" ]; then
   case "$REPLY" in
       y*|Y*)
             for apk in $newapks; do
-              upload=$(curl -s -S "http://www.apkmirror.com/wp-json/apkm/v1/apk_uploadable/$(md5sum "$apk"|cut -f1 -d' ')" | grep uploadable)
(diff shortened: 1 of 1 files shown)


upload_sources.sh
```diff
@@ -186,9 +186,14 @@ if [ -n "$newapks" ]; then
   case "$REPLY" in
       y*|Y*)
             for apk in $newapks; do
-              echo "Uploading $apk to APKmirror.com..."
-              filename="$(basename "$apk")"
-              curl -s -S -X POST -F "fullname=$name (OpenGApps.org)" -F "email=$email" -F "changes=" -F "file=@$apk;filename=$filename" "http://www.apkmirror.com/wp-content/plugins/UploadManager/inc/upload.php" > /dev/null
+              upload=$(curl -s -S "http://www.apkmirror.com/wp-json/apkm/v1/apk_uploadable/$(md5sum "$apk"|cut -f1 -d' ')" | grep uploadable)
```
mfonville: use spaces for the cut parameters:
`cut -f 1 -d ' '`
(to be fully compliant with some older/stricter versions of it)
it is also often more readable to keep space around the pipe

also best to pass the `-q` to grep, to keep it silent if matching, or not; and " around the keyword

PR #422, opened 2017-01-17
Add SoundRecorder to optional remove list

dobragab:


mfonville:
@dobragab thanks!

could you add the keyword to the documentation in the wiki?

NicholasBuse:
I this to the Wiki.

PR #424, opened 2017-01-18
Added option to gapps-config.txt for installing phone stuff on tablet

### LINKED ISSUE #420: Force Google Messenger installation on tablets with SIM card slot/s
I have a tablet, when I flash Open GApps, Google Messenger is not installed because my device is detected as a "tablet".

dobragab:
If you add Phablet keyword to your gapps-config, that will install dialerframework, dialergoogle, and messenger even if device type is a tablet.

Fixes #420.


mfonville:
Thanks, I will take a look at this, when also considering that Hangouts will be dropped as SMS provider; so make sure everything gets consistent :)

osm0sis:
Presumably abandoned.

PR #432, opened 2017-01-24
addons.d: Prepare for busybox-less environment

mdmower:
toybox (used by cm recovery and optional in twrp) lacks 'awk' and has
incomplete implementations of 'find' and 'xargs'.

issue: resolution
* no awk: use cut
* rmdir does not support --ignore-fail-on-non-empty: redirect errors
  to /dev/null (note toybox find does not support -empty, even though
  it claims it does in --help)

### DIFF
--- a/scripts/bkup_tail.sh
+++ b/scripts/bkup_tail.sh
@@ -4,13 +4,13 @@ EOF
 # Backup/Restore using /sdcard if the installed GApps size plus a buffer for other addon.d backups (204800=200MB) is larger than /tmp
 installed_gapps_size_kb=$(grep "^installed_gapps_size_kb" /tmp/gapps.prop | cut -d= -f2)
 if [ ! "$installed_gapps_size_kb" ]; then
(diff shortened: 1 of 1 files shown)


mdmower:
This certainly needs a careful review, but it sure would be nice to reduce dependence on busybox-only tools.

mfonville:
Thanks for your PR, this is indeed an improvement that is desired but hard to spot (and neglected ;-) ), I will review and if necessary adjust a bit to merge it in.

mfonville:
@mdmower quotes can be within quotes if within `$()` and it is best to always quote, if possible, to keep input as sane as possible; strongest recommendation is to always quote unless not possible at all and really relying on globbing.

mdmower:
Ah, I see. I tested most of these commands (manually) at the shell and they're working as expected.

mfonville:
Committed as rebased commits: https://github.com/opengapps/opengapps/commit/60ef4f053c89a4193a2d1447a571fe32cc7b6ce5
https://github.com/opengapps/opengapps/commit/c76386811d6b2521805bce17c05b3223f31ab061

PR #438, opened 2017-02-01
Expand LOS14+ CMSetupWizard removal to CM14+  & add Recorder removal support

osm0sis:
Per my conversation with @mfonville we can expand the LOS CMSetupWizard removal to CM14+.

Also I noticed Recorder popped up in the latest LOS14 weekly, so handle that while we're at it. 😃

### DIFF
--- a/scripts/inc.installer.sh
+++ b/scripts/inc.installer.sh
@@ -1363,10 +1363,9 @@ else
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


osm0sis:
Ah good catch! Removed it by mistake when I fixed the case last minute. Thanks!

Should be good now. 👍

Issue #441, opened 2017-02-04
Issues with OpenGApps when CMSetupWizard present from CM/LineageOS


kronflux:
I do realize there are other issues present in the past regarding CMSetupWizard, however as LineageOS is one of the most popular roms for the majority of current devices, I believe it requires deeper attention.
OpenGApps is essentially the standard for GApps packages now, and it's very unfortunate to not have it work for what is also essentially the standard for custom flashed Roms.

I will also note that this issue has not been specifically noted in the Wiki article related to CMSetupWizard nor do I see any open tickets for the issue at the moment.

Thanks!

**Issue description:**

During initial setup the user is asked to transfer data from another device, however there is no way to skip, and you end up in a loop.
This is only an issues with OpenGApps, and is not present in various other GApps packages including(but not limited to) BaNkS Dynamic GApps.

**Reproduction Steps:**

1.  Clean install LineageOS
2.  Clean install OpenGApps
3.  Run through setup process

**Expected behavior:**

When you hit skip, setup should continue.

As noted, this is only with OpenGApps so one assumes(without the technical knowledge to investigate this further) it is a problem with the end resultant structure/configuration after flashing OpenGApps.

**Workarounds:**

1.  Remove CMSetupWizard, which causes other unexpected behaviors, but does solve this particular issue.
2.  Click skip and back repeatedly until either the app crashes(and setup simply continues) or you manage to skip all the way to the home screen where you can then set up manually from the settings menu.
3.  Flash your rom and complete setup before flashing OpenGApps, then reboot and flash it
4.  Actually perform a device data transfer from an old android device

marcthe12:
same here. If do get pass the cm wizrd somehow. Google playservices crash espicaly when you need to sign in. this could mean work around 2 is proven not to work on moto g2. I will try 3 although that goes my g camara.

josephcsible:
The latest versions of OpenGApps remove CMSetupWizard from CM14+ for this very reason, and the wiki page about issues from removing it only applies to CM13 and below. Given that, what exactly is the problem?

marcthe12:
Play service is faulty. It crashes with google sign in. Lineage os Feb 1 weekly and 3/4 daily open gapps have massive issues  with inital setup (something simple as setup wizard or google sign in dont work). I think Lineage side bug has been fixed but the gapps side on does not work with a down grade of lineage. The current work around is downgrade version of both zips before 1st of feburary

josephcsible:
[...]

PR #444, opened 2017-02-09
Fixed typo causing execution of arthmatic opperation

marcthe12:
Fix for #441

### DIFF
--- a/scripts/bkup_tail.sh
+++ b/scripts/bkup_tail.sh
@@ -4,7 +4,7 @@ EOF
 # Backup/Restore using /sdcard if the installed GApps size plus a buffer for other addon.d backups (204800=200MB) is larger than /tmp
 installed_gapps_size_kb=$(grep "^installed_gapps_size_kb" /tmp/gapps.prop | cut -d '=' -f 2)
 if [ ! "$installed_gapps_size_kb" ]; then
-  installed_gapps_size_kb="$(cd /system; size=0; for n in $(du -ak $(list_files) | cut -f 1); do ((size+=n)); done; echo "$size")"
+  installed_gapps_size_kb="$(cd /system; size=0; for n in $(du -ak $(list_files) | cut -f 1); do size=$((size+n)); done; echo "$size")"
   echo "installed_gapps_size_kb=$installed_gapps_size_kb" >> /tmp/gapps.prop
 fi
 


mdmower:
I don't think this is correct. The `(( ))` is intended for += arithmetic. For example, open a mksh shell and test each of these; the first works, the second incorrectly tries to treat the contents of `(( ))` as a command.

Good: `size=0; for n in 1 1 1 1; do ((size+=n)); done; echo $size`
Bad: `size=0; for n in 1 1 1 1; do $((size+=n)); done; echo $size`

marcthe12:
@mdmower Lineage os installer thinks to ((size+=n)) is a command and therefore crashes trying to executed it.
$((size+=n)) works for the installer

mdmower:
Lineage os installer? I'm not familiar with such a thing.

Maybe what you're suggesting is `installer.sh` in the opengapps package is causing `/system/addon.d/70-gapps.sh` to be mangled upon installation? Would you mind pasting a copy of your `/system/addon.d/70-gapps.sh` so I can see if this command does indeed get incorrectly written?

marcthe12:
That is in a comment in linked issue above. Also dash also causes the same error.

marcthe12:
[Here is for more difference](http://unix.stackexchange.com/a/149916)

mdmower:
I must be missing something here. My understanding is that `(( ))` and `$(( ))` both perform the same arithmetic. All I interpret your `$`-sign to be doing is to attempt to run the result of the arithmetic expansion as a command (notice there is no `echo` or other command in front of it).

marcthe12:
It is a bashism. Not all shell support (( )) syntax. Strictly POSIX compliant shells like dash cause the same error as the installer (Maybe recovery and installer uses dash or ash).  $(( )) on the other work in dash, ash, ksh and bash

mdmower:
Ok, I just tested this in `busybox sh`, and I see that it does try to run size+=n as a command. This patch is still incorrect, though, as the output of `$(( ))` (a number) is also run as a command. So a more portable solution would be to use `$(( ))` assigned to a dummy variable.

mdmower:
In fact, this would probably work best; effectively kill the `+=`: `size=$((size+n))`

marcthe12:
Fixed

mdmower:
This is a good fix. You could probably squash the commits, though.

marcthe12:
Embarasing but I am not sure how to, I think I need write acess to this repo to do that

mdmower:
You can force push the squashed commit to your own repository's `patch-1` branch and it'll automatically pull in the update here.

marcthe12:
ready to merge

mfonville:
Thanks guys, your work is much appreciated :-)

PR #445, opened 2017-02-09
addons.d: Simplify free /tmp space variable creation

mdmower:
@osm0sis suggested that older toybox versions might not handle `sed -e` correctly, causing garbage to be assigned to free_tmp_size_kb. We can rely on a shell's native field splitting to "squeeze" the spaces between columns in `df -k`. Simply omit the quotes around $(df ...).

I've tested this in `mksh` and `busybox sh`.


mfonville:
Very good thinking for this trick, the incomplete toybox gives really headaches, with limited `df`, broken `sed`, missing `tr`, no `awk` etc... So we have to be creative

PR #450, opened 2017-02-16
inc.installer.sh: Add LineageSetupWizard

AlexLartsev19:
based on https://github.com/LineageOS/android_vendor_cm/commit/7674e0521af98548bec9ac85c084ce4340329c1c

@mfonville , merge this change, if stock LineageOS SetupWizard will be crash on LOS builds after 2016.02.16. If not - close pull request


osm0sis:
I don't see how to add a review like mention-bot suggests, but the change looks fine. Only question would be whether @mfonville might want to keep CMSetupWizard and LSetupWizard separate.

mfonville:
Thanks!

PR #463, opened 2017-03-14
installer: only take the first instance of a prop

invisiblek:
* If a prop exists more than once in a file that's parsed
  for it, only return the first instance. Otherwise you end
  up with another, line-breaked, instance of said prop, which
  doesn't tend to play nicely with scripts that are expecting
  a single value.

Fixes #.

Changes:
-
-
-

### DIFF
--- a/scripts/inc.installer.sh
+++ b/scripts/inc.installer.sh
@@ -877,7 +877,7 @@ get_fallback_arch(){
 }
 
 get_file_prop() {
-  grep "^$2=" "$1" | cut -d= -f2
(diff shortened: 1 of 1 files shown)


invisiblek:
yea yea, people should be smart enough to not have duplicates, but better to be defensive
i banged my head on the wall over this for far too long to not at least try to prevent anyone else from having to live through it

mfonville:
It is indeed against the `prop`-spec to have a field defined more than once, but I see the practical reason. Thanks!

PR #469, opened 2017-04-01
Added omniswitch to the optional_aosp_remove_list

donkers:
Looks like there was an omniswitch_list created  but it was never added to the optional_aosp_remove_list.


mfonville:
Thanks for spotting and fixing this!

PR #476, opened 2017-04-12
installer: Check /tmp/install for gapps-config

TheCrazyLex:
This will be a popular location for ROMs to place
gapps-config files in.


mfonville:
This a very good and practical idea :)

PR #482, opened 2017-04-21
inc.buildtarget.sh: Only remove packages that are part of the variant…

bloodl0rd:
… being installed

Fix for https://github.com/opengapps/opengapps/issues/324

Fixes #.

Changes:
-
-
-


mfonville:
We do have this behaviour for the regular packages on purpose (it is something that has always been that way since PA GApps).
I.e. a PR that would allow to switch behaviours, would be ok.

osm0sis:
Presumably abandoned.

PR #483, opened 2017-04-21
installer.sh: add oneplus dialer to removal list

Nevax07:
OpenGapps will be able to remove OnePlus Dialer (OxygenOS) to avoid conflicts with Google Dialer.


mfonville:
Thanks!

PR #492, opened 2017-05-12
using signapk.sh

ditta95aR:
Fixes #.

Changes:
-
-
-


ditta95aR:
#492

mfonville:
I was never able to get signapk.sh properly working, there where still some differences in the cryptography that my script outputted compared to the official tool.
We at least now have SignApk from @topjohnwu that works reliable

PR #508, opened 2017-06-24
Remove fugu specific pairing apps

webgeek1234:
Changes:
- Removes the BT pairing apps that are specific to fugu

GamepadPairingService specifically causes a problem on Lineage because it requests a shared id with the main Bluetooth app. That causes a signature check error because gamepad is signed with googles key and Bluetooth is signed with test or lineage keys. End result is no bluetooth at all.

Since these are fugu specific, I'd like to see them removed from the general tvstock package. Aosp pairing works fine on the Shield Android TV.

### DIFF
--- a/scripts/inc.buildtarget.sh
+++ b/scripts/inc.buildtarget.sh
@@ -111,7 +111,6 @@ tvvending"
 
 gappstvstock="backdrop
 castreceiver
-gamepadpairing
 globalkey
 livechannels
 overscan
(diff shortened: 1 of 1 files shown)


webgeek1234:
@mfonville Can you take a look at this, please? I'm holding ROM builds waiting on this issue to be resolved.

mfonville:
Yes, thanks! Will the ROM builders make sure that they then include these apps vendor-style for Fugu?

PR #510, opened 2017-07-02
Strip binaries

arter97:
Especially for tar, those binaries are shipped to the zip files.
Reducing the file size for these are important.


TheCrazyLex:
@mfonville Take a look at this please :)

mfonville:
Thanks for the PR, could you explain to me how you did strip the binaries and the lib? So that we could reproduce it?

arter97:
find . -type f -exec grep {} \; | grep "not stripped" | grep -v "relocatable"
will give you which files are not stripped.
And you can just do "strip" to the following files.
(e.g. arm-linux-androideabi-strip /path/to/file)

mfonville:
thanks!

mfonville:
@yashdsaraf fyi the stripping of the binaries

PR #539, opened 2017-10-05
fix typo so installer.sh exits after install

aclegg2011:
Fixes #.

Changes:
-
-
-


mfonville:
This is not a typo but a function call to line 782

1loudsvt:
Actually it is failing at logcat -d -f "$TMP/logs/logcat"  in the exxit function

jamuir:
looks like this can be closed.

PR #545, opened 2017-10-08
build_gapps.sh: Call api26hack

zhaofengli:
`AndroidPlatformServices` doesn't get added to 8.0+ builds, as `api26hack` (added in a93b82d97e2693e333aac751f906165249739298) isn't actually called.


mfonville:
Oh yeah. Thanks for spotting my error. I am still not at 100% :-/

PR #559, opened 2017-12-08
Remove setupwraith from tvstock

webgeek1234:
SetupWraith requires read and write gsettings permissions which
are only allowed to platform signed apps. Otherwise the wizard will
crash at the end of the loop before marking the device provisioned,
which causes a plethora of issues. So, ROMs need to ship ship
SetupWraith directly, signing it with their platform key.

Lineage is already making changes to work with this issue. But all ROMs will have this same problem if SetupWraith isn't platform signed. Ciwrl compared keys on stock Fugu and found that SetupWraith is platform signed there, and that it's only gapp that is.

### DIFF
--- a/scripts/inc.buildtarget.sh
+++ b/scripts/inc.buildtarget.sh
@@ -102,7 +102,6 @@ googlebackuptransport
 googlecontactssync
 gsfcore
 notouch
-setupwraith
 tvetc
 tvframework
 tvgmscore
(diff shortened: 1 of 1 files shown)


webgeek1234:
@mfonville Can you please look at this?

mfonville:
Thanks for figuring this out!
@NicholasBuse could you drop this app from our sources?

NicholasBuse:
Done: https://github.com/opengapps/all/commit/80129fe0caf072caba746105504f0fff8d7aa350

Aaahh:
Well there's also another hackish way of whitelisting the app in the rom, which could be implemented by the rom dev or by opengapps if opengapps decompiles the framework (system i think) and inserts it in place of the current sign check code, which hasn't been touched in a long time last time i checked (granted 3 years ago), in the zip.

PR #583, opened 2018-03-01
Bundle Carrier Services, Restrict PlatformServices, PackageInstallerGoogle only for API23/26/27, Remove useless dialer_experience.xml

SpasilliumNexus:
-Added Carrier Services (com.google.android.ims) to `gappssuper`. Keyword `carrierservices`. This app is found inside of the 8.1 image for the Pixel 2 under `/system/priv-app/CarrierServices`

-Restricted AndroidPlatformServices (com.google.android.gms.policy_sidecar_o) to only API26, as that app is found only in an 8.0 image for the Pixel 2

-PackageInstallerGoogle will **only** install on API23, API26, and API27. If API24 or API25 is targeted for builds, PackageInstallerGoogle is skipped

-It has been reported that dialer_experience.xml is serving no purpose as of now. Remove it from being included in builds (See https://github.com/opengapps/opengapps/pull/580#issuecomment-369463327 by @elmkzgirxp)

PR #584, opened 2018-03-01
Include dialer_experience.xml in builds for Nougat and below

SpasilliumNexus:
- Found in /system/etc/sysconfig, this file is no longer in builds for Oreo


SpasilliumNexus:
As suggested by @elmkzgirxp and @NicholasBuse in my previous pull here https://github.com/opengapps/opengapps/pull/583

SpasilliumNexus:
Ah so THAT'S what it's for... May I suggest reverting this @NicholasBuse?

SpasilliumNexus:
https://github.com/opengapps/opengapps/pull/585

elmkzgirxp:
That makes sense. It might not have appeared for me since I'm using Pixel/Nexus sysconfigs and changed the device name.

PR #586, opened 2018-03-01
Don't use elif statement for separation

SpasilliumNexus:
Using it in this situation causes cameragooglelegacy-arm, cameragooglelegacy-common, dialergoogle, and pixellauncher to be ignored when building for API23 and API26+, while attempting to bring compatibility with PackageManagerGoogle installation for those APIs. All other APIs build the above mentioned packages fine.


SpasilliumNexus:
@NicholasBuse last pull.

I need a break from this screen...

Issue #590, opened 2018-03-03
Carrier Services Doesn't Get Installed


VincentJoshuaET:
Here's the Aroma GApps config I saved after I flashed and installed the zip:
Messenger=1 BatteryUsage=1 Books=1 CalculatorGoogle=1 CarrierServices=1 CalendarGoogle=1 CalSync=0 CameraGoogle=0 Chrome=1 ClockGoogle=1 CloudPrint=1 ContactsGoogle=1 DialerFramework=1 DialerGoogle=1 DMAgent=0 Docs=0 Drive=1 Duo=0 Earth=1 ExchangeGoogle=1 FaceDetect=1 FaceUnlock=1 Fitness=0 GCS=1 Gmail=1 GoogleNow=0 GooglePay=1 GooglePlus=1 GoogleTTS=1 Hangouts=0 Indic=0 Japanese=0 Keep=1 KeyboardGoogle=1 Korean=0 Maps=1 Movies=1 Music=1 NewsStand=1 NewsWidget=1 PackageInstallerGoogle=1 Pinyin=0 PixelIcons=1 PixelLauncher=1 Photos=1 PlayGames=1 PrintServiceGoogle=1 ProjectFi=0 Sheets=0 Slides=0 Search=1 Speech=1 StorageManagerGoogle=1 Street=1 TagGoogle=1 Talkback=1 Translate=1 VRService=1 Wallpapers=1 WebViewGoogle=1 YouTube=1 Zhuyin=0 inclorexcl=1 

Here's the user's gapps-config from the Open GApps log after installation:
[...]

PR #592, opened 2018-03-03
Update the aroma installer to be inline with the wiki and other fixes

SpasilliumNexus:
This includes:
- Rearranged entries to match the ordering and options given in the wiki: https://github.com/opengapps/opengapps/wiki/Advanced-Features-and-Options
- Adding new selections that were never present before

Other fixes:
- Add "logcat" keyword to "Optional Stock/AOSP/ROM Removal List"


SpasilliumNexus:
This will also solve the issue posted here https://github.com/opengapps/opengapps/issues/590

Issue #591, opened 2018-03-03
Google Package Installer can now be installed in Lineage OS 15.1


VincentJoshuaET:
cmcompatibilityhacks="false" # test for CM/Lineage since they do weird AOSP-breaking changes to their code, breaking some GApps case "$(get_prop "ro.build.flavor")" in cm_*|lineage_*) cmcompatibilityhacks="true"; if [ "$rom_build_sdk" -ge "24" ]; then aosp_remove_list="${aosp_remove_list}cmsetupwizard"$'\n';fi;; # CMSetupWizard is broken in LineageOS 14+ and can be safely removed on CM14+ as well esac

|lineage_* can now be removed as Google Package Installer is now working properly on Lineage OS 15.1 builds.

SpasilliumNexus:
Currently looking into this.

SpasilliumNexus:
Same thing with this one. Thanks for reporting the issues.

PR #593, opened 2018-03-03
Allow PackageInstallerGoogle installation on LineageOS 15.1

SpasilliumNexus:
For now, the restriction is lifted for Lineage 15.1, as I cannot verify if PackageInstallerGoogle will also work with LineageOS 15.0.


SpasilliumNexus:
As suggested here by @VincentJoshuaET https://github.com/opengapps/opengapps/issues/591

PR #608, opened 2018-03-23
O ATV Updates

webgeek1234:
First commit updates the scripts to use the new tvlauncher for api 26+.
Second commit fixes tvstock using the incorrect music app.
Third commit adds back setupwraith for api 26+. This has been tested against Lineage 15.1. It's possible the apk from fugu 8.0 also fixes the issues on api <=25, but this has not been tested, so this commit only adds it in for newer apis.
Fourth commit removes a few more ATV apps. These error out on mismatched shared user sigs, making them unusable unless platform signed.

Fugu is now having space issues. I'll follow up later with a PR to add a tvmini variant. This is mainly to get arm64 builds in shape for foster (nvidia shield android tv).

### DIFF
--- a/scripts/inc.buildtarget.sh
+++ b/scripts/inc.buildtarget.sh
@@ -97,8 +97,7 @@ launcher
 mms
 picotts"
 
-gappstvcore="bugreport
-configupdater
+gappstvcore="configupdater
 googlebackuptransport
 googlecontactssync
 gsfcore
@@ -110,11 +109,9 @@ tvvending"
 
 gappstvstock="backdrop
 castreceiver
-globalkey
 leanbacklauncher
 livechannels
 overscan
-remotecontrol
 secondscreensetup
 secondscreenauthbridge
 talkback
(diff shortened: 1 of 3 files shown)


webgeek1234:
@mfonville Bump. Looking to ship Lineage 15.1 for foster next week. Can this and the related all PR get reviewed, please?

NicholasBuse:
@webgeek1234 I have a couple of questions:

1. I saw you dropped 3 packages in [inc.buildtarget.sh-->get_package_info()](https://github.com/webgeek1234/opengapps/blob/e906e332b02eda14c890ba4b031881fb59d32af1/scripts/inc.buildtarget.sh#L208) (not including renames): bugreport, globalkey, and remotecontrol.  Some of those packages were dropped from priv-app targeting in [inc.sourcehelper.sh-->getapkproperties()](https://github.com/webgeek1234/opengapps/blob/e906e332b02eda14c890ba4b031881fb59d32af1/scripts/inc.sourceshelper.sh#L26). Are these are meant to be moved to app or dropped from the APK repo?

1. I do not have much knowledge of the tv builds, but it seems that you intended to keep legacy builds working (e.g. (tv|leanback)(launcher|recommendations) rename/extension). I checked and assume the dropped packages (above) simply were not used in older builds. Correct?

webgeek1234:
They are not usable on any api level as mentioned in the explanation in the PR comment. They give shared user signature mismatches, meaning they must be platform signed to run. As is, they silently error (to the user, logcat lists what's going on) and do nothing.

Yes, I'm trying to keep backwards compatibility. I renamed launcher and recommendations to try and avoid confusion with the names. The new launcher is named tvlauncher, the same as opengapps previous internal name for the launcher.

NicholasBuse:
Ok, then you agree that I can drop the following from the APK repo:

1. com.google.android.tv.bugreportsender
2. com.google.tungsten.bugreportsender
3. com.google.android.athome.globalkeyinterceptor
4. com.google.android.athome.remotecontrol
5. com.google.android.atv.customization (actually not there...)

You don't have update the ALL PR if you don't want, I can cleanup after merging...

webgeek1234:
Affirmative. And customization was removed in a previous PR, but I noticed I missed a reference, so I cleaned it up here.

NicholasBuse:
Great! Thanks for the PR and the information.

Issue #44, opened 2015-07-05
signapk.sh does not build a verifiable signature


Feelivent:
Verifying the zip signature in TWRP 2.8.7.0 fails. 
Tested with open_gapps-arm-4.4-stock-20150704.zip and open_gapps-arm-4.4-stock-20150701.zip

Please do sign, even if it's just with AOSP test-keys, so at least integrity is guaranteed.
http://wiki.rootzwiki.com/Signing

mfonville:
It is a known issue. We are still trying to work out the last problems with the signature, because we can't use apksign.jar and we do this manually from the shell. The signature in binary form in the zip comment field is still broken for unknown reason :-/

mfonville:
I switched the stable branch for now to using signapk.jar, so from tomorrow on the zips will be signed properly.
We still want to fix the shell version though...

Lekensteyn:
These implementations might be of interest:

update.zip installer and signature verification:
 - https://github.com/CyanogenMod/android_bootable_recovery/blob/cm-12.1/install.cpp#L249
 - https://github.com/CyanogenMod/android_bootable_recovery/blob/7b0fd22410453162a44d0fba9e446acb0dd904e9/verifier.cpp#L107

SignApk source:
 - https://github.com/android/platform_build/blob/master/tools/signapk/SignApk.java

osm0sis:
Is this working yet? Signing a shell update-binary zip always created issues with unzip until I started using Chainfire's zipadjust binary. I built it for Windows but it should be easy to build for your build server as well.

http://forum.xda-developers.com/android/software-hacking/dev-complete-shell-script-flashable-zip-t2934449/post56621542

mfonville:
No, it doesn't work yet. I was very close to it, but at some point other more important developments took priority.
Preference is still to get it working using (posix) shell tools, if possible.

osm0sis:
Definitely check out Chainfire's source for zipadjust and MinSignApk then, maybe you can replicate what they do with shell tools. :+1:

nezorflame:
@mfonville this should be closed now.

PR #611, opened 2018-03-26
Switch from SignApk to ZipSigner 2.1 from @topjohnwu

osm0sis:
https://github.com/topjohnwu/Magisk/tree/master/utils/src/main/java/com/topjohnwu/utils
- signs EDIFY and shell-based zips to both unpack and verify correctly without needing a further zipadjust/MinSignApk
- streams the file input while signing so the memory requirements are minimized (e.g. A 1069mb Open GApps ARM 7.1 AROMA zip was able to be signed on a WinXP VM with 1.0gb of memory!)
- has testkeys built in when no others are supplied on the command line so is completely portable


osm0sis:
This might also fix #44. Not sure. 😃

NicholasBuse:
@mfonville is happy, thanks @osm0sis !

PR #642, opened 2018-06-24
Updated AROMA installer to 3.0.0

L2jLiga:
Fixes #.

Changes:
- Updated AROMA installer to version 3.0.0


L2jLiga:
I tested this patch on few devices, seems like all fine. (actually I agree that we need to test on more devices)
About beta label.. looks like AROMA is discounted..

MartinX3:
Yes, aroma is discontinued.

MartinX3:
Btw. I think they build in aroma support into TWRP for the newer devices which get freezes.

PR #666, opened 2018-09-18
Add storagemanagerstock to AOSP remove list

injust:
Looks like this was missing, `StorageManagerStock` is currently a no-op.


NicholasBuse:
Good catch! Thanks!

PR #676, opened 2018-10-18
inc.installer.sh: getprop from vendor as well

ThEMarD:
* treble ROM builds move a lot of properties to /vendor/build.prop for compatibility with GSI's
* without this change, OpenGapps would fail to get the values of certain props such as DPI or camera API version (which would tell OpenGapps that Google Camera version = Legacy)

### DIFF
--- a/scripts/inc.installer.sh
+++ b/scripts/inc.installer.sh
@@ -711,16 +711,18 @@ if [ -n "$(cat /proc/cmdline | grep slot_suffix)" ];
 then
   device_abpartition=true
   SYSTEM=/system/system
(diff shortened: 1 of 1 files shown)


ThEMarD:
I'm still learning software development, so I'm unsure if this is the best way to go about implementing this, so if someone could please review it or even improve upon it? That would be appreciated.

ThEMarD:
Also, I don't have a device with a/b partitions, so if someone that does have a device with it? Testing this commit would be really appreciated.

NicholasBuse:
Thanks for the PR (and for following up)

Issue #626, opened 2018-05-01
Aroma doesn't remove the LineageOS Recorder app


grezzaa:
The last couple of times I've used the Aroma installer on my phone. The LineageOS Recorder app has not been removed even though I selected the option to remove it. 

Not sure if relevant but I'm using a Redmi Note 4X Mido with whatever the latest Lineage and Gapps are at the time. This has been happening for at least a month or so.

PR #678, opened 2018-10-19
Fix removal of LineageOS Recorder

Peter-Levine:
Fixes https://github.com/opengapps/opengapps/issues/626

It appears that in commit https://github.com/opengapps/opengapps/commit/a30207bd382092986b18e24dd3213aef530b4cec, the actual code to add LineageOS Recorder to removal list if `LRecorder=1` is found in rem.prop was inadvertently omitted.

### DIFF
--- a/scripts/inc.aromadata.sh
+++ b/scripts/inc.aromadata.sh
@@ -825,6 +825,12 @@ then
 endif;
 
 if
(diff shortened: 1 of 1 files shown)


SpasilliumNexus:
You're right. I've looked over having that left in. Good catch!

PR #685, opened 2018-10-31
Removal of Pixel's Setup Wizard in favor of "Universal" variants

SpasilliumNexus:
**Scripts: Remove Data Transfer Tool from gappscore**
Using a non-Pixel SetupWizard, we do not need to have this included in installation. The new SetupWizard will not force itself to download the tool either, and usually ending up being stuck in setup after. Notifications (such as the download notification) will also work properly now, instead of having to disable the tool after setup.

- Moved `datatransfertool` from "`core`" to "`gapps`", and only removed the keyword, in case that app is needed for something else.

- I have included screenshots of the alternative SetupWizard [in an album here](https://imgur.com/a/QT2S0kA), which shows all of the screens a user goes through.

- In the Android Settings menu, there is no trace of the "Pixel" string, for example "Finish setting up your Pixel phone", but instead, "Finish setting up your device".

- The last two screenshots in the album show that the download notification works again without having to disable or uninstall Data Transfer Tool.

- The source of these apps were taken from APKMirror:
Oreo: [Setup Wizard 227.4773854](https://www.apkmirror.com/apk/google-inc/setup-wizard/setup-wizard-227-4773854-release/)
Pie: [Setup Wizard 228.4976421](https://www.apkmirror.com/apk/google-inc/setup-wizard/setup-wizard-228-4976421-release/)

**Scripts: Remove libbarhopper library from installation**
Not needed anymore with the alternative SetupWizard

- These alternative wizards do not include a library for ARM64 devices like the Pixel's SetupWizard does. After testing around nearly every part of this alternative Setup Wizard, no warnings or errors relating to these were found in logcats.

### DIFF
--- a/scripts/inc.buildtarget.sh
+++ b/scripts/inc.buildtarget.sh
@@ -252,10 +252,7 @@ get_package_info(){
                                 packagelibs="libsketchology_native.so"
                               fi;;
     setupwizard)              packagetype="Core"; packagename="com.google.android.setupwizard"; packagetarget="priv-app/SetupWizard";; #KitKat only
-    setupwizarddefault)       packagetype="Core"; packagename="com.google.android.setupwizard.default"; packagetarget="priv-app/SetupWizard";
-                              if [ "$API" -ge "28" ] && [ "$ARCH" = "arm64" ]; then  # On Android 9.0 there is now an ARM64 library included with SetupWizard. Not required, but here for completeness.
-                                packagelibs="libbarhopper.so"
-                              fi;;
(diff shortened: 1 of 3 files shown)


NicholasBuse:
Thanks @SpasilliumNexus !  All three looked good and are merged

patmac3:
Since the switch to the new universal setup wizard, all my Google cloud backups fail to restore.

SpasilliumNexus:
@patmac3 
It's been addressed: https://github.com/opengapps/opengapps/pull/690
Update to the November 13th build.

PR #694, opened 2018-11-17
Add missing permission files for atv

webgeek1234:
Changes:
- Add missing permission files for atv

Mainly to get cast receiver working again, but probably fixes a lot of other, more subtle stuff too.

Edit: Needs opengapps/all/#40.


NicholasBuse:
Thanks!

PR #695, opened 2018-11-17
Fix tmp directory typo

### LINKED ISSUE #669: The addon script backs up files into /tmp so that Lineage OS installation fails
**Issue description:**

Open-Gapps' addon script backs up files into /tmp. So during Lineage OS installation, /tmp is filled by gapps backup and Lineage OS installation fails. Trying 20180904, I think that older versions than 20180904 are OK.

**Reproduction Steps:**

1. Install latest open-gapps super set.
2. Install Lineage OS

**Expected behavior:**

Lineage OS installation does not fail. Open-Gapps addon script should back up files into sdcard.

**Observed behavior:**

/tmp/backupdir is filled by /system backup of over 800MB. There is no space in /tmp.

luca020400:
Fixes #669

### DIFF
--- a/scripts/inc.installer.sh
+++ b/scripts/inc.installer.sh
@@ -2436,7 +2436,7 @@ for reqdapp_name in $reqd_list; do
 done;
 
 # Create final addon.d script in system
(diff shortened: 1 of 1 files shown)


ThEMarD:
Works fine on my LeEco Le Max 2 (arm64-28-nano)

PR #696, opened 2018-11-26
Allow Markup on ARM, Search more locations for gapps-config, prevent stock AOSP Dialer from being removed by default

SpasilliumNexus:
**Scripts: Extend Google Markup to ARM devices**
We now have an ARM variant of the library needed for Markup to work on more than just ARM64 devices. The origin of this library is from Pixel Experience [here](https://github.com/PixelExperience/vendor_pixelstyle/tree/pie/lib). Multiple testers in the unofficial OpenGApps Telegram group have confirmed with me that Markup now works perfectly. Unfortunately, I could not find an ARM factory image for Pie to source it from.

**Installer: Add support for locating gapps-config on the root of /sdcard**
In addition to the original location of finding a gapps-config file in `/sdcard/Open-GApps/`, let's also search the root of `/sdcard` for a gapps-config too.

**Prevent stock AOSP Dialer (Phone) from being removed**
Remove the entry for "priv-app/Dialer", which will keep it, and ignore any option in the config relating to it.

### DIFF
--- a/scripts/inc.aromadata.sh
+++ b/scripts/inc.aromadata.sh
@@ -168,7 +168,7 @@ form(
       "KeyboardGoogle",     "<b>Gboard (Google Keyboard)</b>",       "",                      "check",
       "Korean",     "<b>Google Korean Input</b>",       "",                      "check",
       "Maps",     "<b>Google Maps</b>",       "",                      "check",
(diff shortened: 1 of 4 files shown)


SpasilliumNexus:
Related pull for Markup on ARM: https://github.com/opengapps/arm/pull/11

SpasilliumNexus:
Resolves https://github.com/opengapps/opengapps/issues/672

PR #705, opened 2018-12-16
Unbreak system-as-root A-only support

luk1337:
* When backuptool.sh is ran during ROM updates
  it sets S variable to "/system/system", so we can
  use it instead of hardcoding "/system". That way
  proper permissions are restored properly.


luk1337:
See https://review.lineageos.org/q/topic:backuptool-sar-a as the reference.

luk1337:
@NicholasBuse bump?

PR #716, opened 2019-01-28
Android Pie requires system-as-root

hecz0r:
just change /system to /system/system. otherwise, the gapps wouldnt flash on Lineage 16.

Fixes #.

Changes:
-
-
-


nezorflame:
This should be closed in favour of https://github.com/opengapps/opengapps/pull/761

PR #736, opened 2019-04-24
inc.installer: Properly check for A-only system-as-root devices

DD3Boh:
* Checking /proc/mounts will work only if /system_root is mounted in that exact moment, while this will work unconditionally in TWRP

### DIFF
--- a/scripts/inc.installer.sh
+++ b/scripts/inc.installer.sh
@@ -715,7 +715,7 @@ then
   device_abpartition=true
   SYSTEM=/system/system
   VENDOR=/vendor/vendor
-elif [ -n "$(cat /proc/mounts | grep /system_root)" ];
+elif [ -n "$(cat /etc/fstab | grep /system_root)" ];
 then
   device_abpartition=true
   SYSTEM=/system_root/system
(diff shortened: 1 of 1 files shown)


DD3Boh:
DO NOT MERGE (yet)
Still needs a review for Berkeley that right now is mounting /system weirdly for A-only system-as-root

randomblame:
Ok fixed it, new twrp, with the new layout, pulled your changes into opengapps and tested and it works like a charm thanks.

scripts/inc.installer.sh
```diff
@@ -713,16 +713,19 @@ nogooglewebview_removal_msg="NOTE: The Stock/AOSP WebView is not available on yo
 if [ -n "$(cat /proc/cmdline | grep slot_suffix)" ];
 then
   device_abpartition=true
-  SYSTEM=/system/system
+  SYSTEM_MOUNT=/system
+  SYSTEM=$SYSTEM_MOUNT/system
   VENDOR=/vendor/vendor
-elif [ -n "$(cat /proc/mounts | grep /system_root)" ];
+elif [ -n "$(cat /etc/fstab | grep /system_root)" ];
```
mfonville: Why isn't the mount available in proc? that should be more reliable than a fstab (because it checks if the kernel has actually mounted this location

DD3Boh: It's because /proc/mounts shows only the partitions currently mounted, so if system isn't, only fstab will contain the path
So it's better to use /etc/fstab because /proc/mounts just doesn't work

mfonville: Alright, I understand!

PR #758, opened 2019-08-18
installer.sh: Fix installation for A/B devices

aviraxp:
A/B devices are compulsory system-as-root devices. In a properly done
recovery such as Lineage Recovery, system is mounted as /system_root.

Just try /system_root first, and keep the old slot detecting as fallback.


aviraxp:
In addition, this should not break incorrectly built recoveries as there is still a fallback.

PR #761, opened 2019-08-22
Fix SAR on Lineage Recovery

luca020400:
Unlike TWRP Lineage Recovery mounts system_root in /system_root ( as expected by Android )


luca020400:
First of all the old logic is kept intact when it comes to mounting
I fixed the A/B and SAR detection and added a fallback usecase to use /dev block directly as mount in case the mount point isn't specified in fstab ( always the case for SAR )

luca020400:
@mfonville Mind taking a look?

fawazahmed0:
https://github.com/opengapps/opengapps/issues/776

PR #787, opened 2019-12-25
Copy logs to sdcard/ when using sideload

ipdev99:
When using adb sideload, log files are not saved.
(/sideload is RO)

If you use a gapps-config file, the logs will be saved in the same directory as the config file.
If you do not use a config file, then the log files are copied to /sideload and disappear.

For now, copying the logs to /sdcard when /sideload is detected seems to work.


ipdev99:
Closed for now.
Had to clean up my repository.
Will resubmit.

PR #797, opened 2020-01-24
system-as-root fixes

dianlujitao:


osm0sis:
It would be better to test the actual mount to see whether it's SAR or not rather than rely on the prop which may or may not be set correctly in recovery.

Also, rm -rf should never be pointed at system because the umount -l (lazy umount) may or may not have completed, as I mentioned in your Magisk pull request, but it's even more serious with GApps since system is mounted rw. So please change these to rmdir.

dianlujitao:
> It would be better to test the actual mount to see whether it's SAR or not rather than rely on the prop which may or may not be set correctly in recovery. Usually we look for /system/system/build.prop in such tests.

It's currently the strategy of magisk and I do agree it's better, but I'm on vacation now and don't have time to implement, it would be nice if you could do it and send another PR.

> Also, rm -rf should never be pointed at system because the umount -l (lazy umount) may or may not have completed, as I mentioned in your Magisk pull request, but it's even more serious with GApps since system is mounted rw. So please change these to rmdir.

Thanks for the heads-up, I changed locally before pushing but forgot to amend, fixed now.

PR #798, opened 2020-01-25
installer: mount and umount instead of mount --move

YoDevil:
* mount --move does not work if the old mount point's parent
  is a shared mount, which it is.

for reference: https://github.com/LineageOS/android_kernel_lge_msm8974/blob/lineage-17.1/fs/namespace.c#L1810

this will prevent always falling back to the mount /dev/block/bootdevice option

### DIFF
--- a/scripts/inc.installer.sh
+++ b/scripts/inc.installer.sh
@@ -1022,7 +1022,7 @@ case $ANDROID_ROOT in
   /system)
     if [ -f /system/system/build.prop ]; then
       setup_mountpoint /system_root;
-      mount --move /system /system_root;
+      mount /system /system_root && umount /system;
       if [ $? != 0 ]; then
         umount /system;
         umount -l /system 2>/dev/null;


nezorflame:
@osm0sis WDYT?

osm0sis:
It isn't shared at that point. The case where --move gets used would only have /system mounted and /system/system/build.prop is detected. The bind mount that would share it doesn't occur until after.

YoDevil:
@osm0sis It does not need to be actually shared in that moment, but to be flagged as a shared mount.
Also it is not "/system" that it checks, it checks for the parent "/", and "/" is a shared mount
```d802:/ # /tmp/busybox-arm cat /proc/1/mountinfo                                                                                           
1 1 0:1 / / rw shared:1 - rootfs rootfs rw,seclabel
[...]
23 1 179:30 / /system ro,relatime shared:12 - ext4 /dev/block/platform/msm_sdcc.1/by-name/system ro,seclabel```

**Comment by osm0sis** [comment]
That premise isn't right then.. GApps aren't (or should never be) flashed from a booted system where / exists as a mountpoint and could be shared, in recovery it's /system_root or /system and / is TWRP's ramdisk.

**Comment by YoDevil** [comment]
Not sure what you mean by "/ does not exist as a mountpoint"?
Anyhow difference seems to be that in Q recovery all mounts are shared, that's where I got the above snippet from.

**Comment by osm0sis** [comment]
That doesn't make sense, there is no "Q recovery", TWRP doesn't even have aosp 10 recovery building yet.

**Comment by YoDevil** [comment]
At @LineageOS we mantain recovery rebased on whatever android release.
Checkout https://github.com/LineageOS/android_bootable_recovery

This issue aroused while testing Lineage 17 recovery against popular zips, such as opengapps.

**Comment by osm0sis** [comment]
Oh so Lineage recovery has a completely different environment from TWRP, and that's where it doesn't work for you.. you probably should have lead with that. :wink:

**Comment by YoDevil** [comment]
Yeah I figured, my bad!

**Comment by osm0sis** [comment]
Does it work with the fallback? Basically sounds like it's the reason I ended up having to include it.

**Comment by YoDevil** [comment]
The fallback works, if the bootdevice symlink is present. But most devices are not currently symlinking /dev/block/bootdevice in recovery as it never was explicitly needed.
Or, some other devices are using a different path, such as https://github.com/LineageOS/android_device_google_marlin/blob/lineage-17.1/init.recovery.common.rc

**Comment by osm0sis** [comment]
I'll look into it. Thanks!

**Comment by osm0sis** [comment]
@YoDevil, can try to look into @LineageOS Recovery making both bootdevice symlinks and symlinks to the current slot for all major slotted partitions like TWRP does? It's way more sensible if you want your recovery to not break every zip in existence.

**Comment by YoDevil** [comment]
I'm looking at trees for devices of my era like hammerhead, bacon or htc m8, and none seems have added a bootdevice symlink but are still receiving official twrp builds.
It does not look like a TWRP "established thing", but just a thing that happens to be around on _modern_ devices.

**Comment by osm0sis** [comment]
We're talking about SAR, on modern devices. Why would hammerhead factor into this at all? TWRP on modern devices makes these symlinks, so should Lineage Recovery on modern devices.

**Comment by YoDevil** [comment]
Android 10 /system is SAR for any device.
For all devices that were not SAR in Pie it uses "2 stage init" to boot to bootimage and then chroot to system

**Comment by osm0sis** [comment]
Correct. Not sure what you're driving at. I'm asking Lineage Recovery to be more conforming for modern devices to help keep zips compatible. You've already done something completely different by actually mounting system to / and running recovery over top of that, so you should at least try to make things easier by adding bootdevice and boot_a -> boot (if slot a is active) symlinks.

I'll do what I can for the zips I have some influence over, but you changing the entire recovery environment and then expecting all zips to jump to fix it is a bit ridiculous.

**Comment by YoDevil** [comment]
Nono, wait, we're not mounting system and running recovery from system. Just aosp recovery changed, a lot, to be more like a whole system. We're following aosp recovery rather than TWRP and that's not gonna change; if anything TWRP should, and i'm sure will, catch up to aosp Q recovery or it will be unable to support dynamically partitioned devices and such.

Here I was arguing that bootdevice symlinks should not be considered a given in recovery (or in system!). TWRP too does not have them for legacy devices that are though still capable of rocking up-to-date android.
I'm sure modern devices will have bootdevice symlinks in Lineage Recovery, since that's what CAF does in their trees as fstab references them. But just like in TWRP, currently legacy devices do not symlink bootdevice, in LineageOS nor in Lineage Recovery.

**Comment by osm0sis** [comment]
Hmm okay! Not so bad! Well, I made my boot.img insecure on my OP7Pro right now on Lineage Recovery and it doesn't appear as drastically different as it seemed like you were saying at first.

```
rootfs on / type rootfs (rw,seclabel)
tmpfs on /dev type tmpfs (rw,seclabel,nosuid,relatime,mode=755)
devpts on /dev/pts type devpts (rw,seclabel,relatime,mode=600,ptmxmode=000)
proc on /proc type proc (rw,relatime,gid=3009,hidepid=2)
sysfs on /sys type sysfs (rw,seclabel,relatime)
selinuxfs on /sys/fs/selinux type selinuxfs (rw,relatime)
tmpfs on /mnt type tmpfs (rw,seclabel,nosuid,nodev,noexec,relatime,mode=755,gid=1000)
none on /acct type cgroup (rw,relatime,cpuacct)
tmpfs on /tmp type tmpfs (rw,seclabel,relatime)
tmpfs on /storage type tmpfs (rw,seclabel,relatime,mode=050,gid=1028)
pstore on /sys/fs/pstore type pstore (rw,seclabel,relatime)
none on /config type configfs (rw,relatime)
adb on /dev/usb-ffs/adb type functionfs (rw,relatime)
```

Then /system_root as a mountpoint for system partition (despite /etc/fstab saying /), and /system as a symlink to /system_root/system (just like many modern TWRP do).

I'll directly test out the scripts via adb shell and see if there's anything more to improve, but I can safely say this PR won't be accepted as-is, since it'll break the use-case of the fallback for devices where mount --move fails; the bind over /system takes place after this, once /system or /system_root was correctly mounted. I might have already fixed this in my AK3 update over here, even, which will be brought over to TWRP soon once I confirm our Lineage issues are resolved: https://github.com/osm0sis/AnyKernel3/commit/062b44c75a967a9d2a603459e1092f09e380e825

Thanks for your help in understanding this mess! :)

osm0sis:
I think this should do the trick: https://github.com/osm0sis/AnyKernel3/commit/03af96a5415cd0b508e542a1ed8c9eb386c322a6

YoDevil:
Not really, the original point of this PR was to fix the issue where:
$ANDROID_ROOT is /system, but that's a SAR partition as we're talking Android 10.
The mount --move to /system_root fails because we're in Lineage Recovery.
We still fall back to the mount bootdevice path and fail that too.
We end up with no system mounted, and we can't install.

osm0sis:
Well this PR isn't a good solution for it either, then, since your changes break other use-cases in existing released recoveries. I don't have a device with Lineage 17.1, so I was testing with the Lineage 16 Recovery.

Lineage 17.x hasn't been officially released yet, correct? Please consider adding bootdevice symlinks as part of the bring-up for it, and while you're at it add symlinks to the active slot partitions like TWRP does.

I still think you could make Lineage Recovery behave better instead of putting the burden on everyone else to rewrite every zip in existence for you.

YoDevil:
--- Addressing the rant ---
I personally already added a symlink to bootdevice in my tree, but other mantainers were not as happy to. But if and when TWRP manages to merge/rebase on top of aosp Q recovery, this script will not work on TWRP too, on those devices that are not symlinking bootdevice.
-----------------------------

But back on the matters of this PR:
I don't see why this breaks existing usecases.
If system is SAR, we will end up with `/system` UNMOUNTED and `/system_root` mounted, so we can procede to bind mount` /system` to `/system_root/system` and install.
If it wasn't SAR, then this PR changes nothing.
In my eyes I'm not changing any surrounding behaviour. Am I missing something?

CaptainThrowback:
> --- Addressing the rant ---
> I personally already added a symlink to bootdevice in my tree, but other mantainers were not as happy to. But if and when TWRP manages to merge/rebase on top of aosp Q recovery, this script will not work on TWRP too, on those devices that are not symlinking bootdevice.
> 
> But back on the matters of this PR:
> I don't see why this breaks existing usecases.
> If system is SAR, we will end up with `/system` UNMOUNTED and `/system_root` mounted, so we can procede to bind mount `/system` to `/system_root/system` and install.
> If it wasn't SAR, then this PR changes nothing.
> 
> I'm just moving the _mount via bootdevice_ fallback from _when mount --move fails_ to _when mount && umount fails_ (which btw I don't see how could happen).
> In my eyes I'm not changing any surrounding behaviour. Am I missing something?

For the record (and since you mentioned M8), I also [added bootdevice symlinks](https://github.com/CaptainThrowback/android_device_htc_m8/blob/android-8.1/recovery/root/init.recovery.qcom.rc#L5) for it in order to have full support for newer AOSP-based ROMs and since I was updating for SAR support for Android 10. I suspect that will be the norm rather than the exception.

osm0sis:
I'm testing Lineage 17 Recovery on my OP7Pro from LuK1773' builds. 👍

Digging into this PR again now that I've got the AOSP Q recovery environment to test within, your change of `mount /system /system_root` simply bind mounts the recovery ramdisk's skeleton /system over top of /system_root, which wouldn't serve a purpose, since our aim is having the actual system partition get mounted. Then because /etc/fstab doesn't contain a /system entry your `umount /system` always fails, which will always trigger the fallback to my bootdevice method anyway.

Lineage 17 Recovery appears to have some fundamental issues currently:
- /etc/fstab's system partition entry has / as the mount point, where it should be either /system or /system_root in a recovery to have working mount commands
- /etc is symlinked to /system/etc, so mounting /system properly will make /etc/fstab inaccessible and break all simple mount commands
- /sbin is either a directory with basically nothing in it (e.g. dashd only on my OP7Pro), or I see you have a commit to symlink it to /system/bin (which only works if /sbin is empty due to the rmdir), so there's no directory with utilities permanently on the PATH independent of mount state, much as with /etc/fstab above, and since all /system/bin utilities from the actual /system/bin require /apex files to also be mounted this results in every recovery shell command being broken once /system gets mounted correctly

So, please consider these issues blocking for most zips that would want to support Lineage 17, and again, please also seriously consider my request to have recovery make /dev/block/bootdevice symlinks for all Lineage 17 recoveries and make symlinks for the active slot partitions, (e.g. /dev/block/.../boot to /dev/block/.../boot_a for active slot _a), since TWRP definitely does the latter, and that will give Lineage 17 Recovery the broadest zip compatibility once it officially ships.

PR #803, opened 2020-01-31
installer: mount and output fixes, touch ups

osm0sis:
- fix case where TWRP has /system_root mountpoint but ROM is non-SAR
- fix regression in ui_print when final character is \
- adjust section spacing for uniformity, and clean up one last trailing ;
- fix the following bugs and/or stderr in the log:
```
grep: /system/product/etc/: No such file or directory
ash: bad number
find: /system/product/app: No such file or directory
find: /system/product/priv-app: No such file or directory

find: /system/product/app: No such file or directory
find: /system/product/priv-app: No such file or directory
find: /system/vendor/pittpatt: No such file or directory

find: -exec requires an argument
```

### DIFF
--- a/scripts/inc.installer.sh
+++ b/scripts/inc.installer.sh
@@ -93,7 +93,7 @@ done
 if [ -e "$bb" ]; then
   install -d "$l"
   for i in $($bb --list); do
-    if ! ln -sf "$bb" "$l/$i" && ! $bb ln -sf "$bb" "$l/$i" && ! $bb ln -f "$bb" "$l/$i" ; then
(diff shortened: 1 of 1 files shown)


scripts/inc.installer.sh
```diff
@@ -1331,7 +1343,7 @@ get_prop() {
   #check known .prop files using get_file_prop
   local propfile propval
   for propfile in $PROPFILES; do
-    test "$propval" && break 2 || propval="$(get_file_prop $propfile $1 2>/dev/null)"
```
mfonville: you sure we don't need to break the loop here twice on purpose?

osm0sis: We don't, break 2 is what has been keeping the device_name for loop from working correctly lately.

sanpunk06:
Hello

PR #805, opened 2020-02-02
Include googlepixelconfig and datatransfertool for pixel devices

razorloves:
The pixel config files were separated from defaultetc in change
4857790 and removed from gapps zips. Pixels need these
so change it to a core package.


razorloves:
oh, i think i need to add the following to the installer script so that it's only installed on pixels.
googlepixelconfig) if [ "$googlepixel_compat = "true" ]; then extract_app "$archive"; fi;;

razorloves:
Added new change to include datatransfertool.  This fixes a really old bug that prevented pixel devices from being able to use opengapps.
Built and tested and working good on both pixel and non-pixel devices.

nezorflame:
LGTM

nezorflame:
@mfonville please merge this.

mfonville:
Thanks!

PR #810, opened 2020-02-21
speedup xz compression by using all available threads

yurikoles:
Changes:
- Use all threads for xz compression.

Signed-off-by: Yurii Kolesnykov <root@yurikoles.com>


mfonville:
Thanks!

PR #814, opened 2020-03-04
Fix package drag n drop installation in Genymotion

JulienBolard:
Some Android devices (nearly all) don't have any /etc/fstab file.
Using 'mount' command will fail when trying to remount partitions automatically.

Just simply ignore these partitions handling when /etc/fstab is not present and
just remount ANDROID_ROOT rw if needed.

Fixes opengapps#813

Changes:
- Keep old mounting behaviour when device don't have any /etc/fstab file.

### DIFF
--- a/scripts/inc.installer.sh
+++ b/scripts/inc.installer.sh
@@ -1108,10 +1108,16 @@ ui_print "- Mounting partitions"
 set_progress 0.01
 test "$ANDROID_ROOT" || ANDROID_ROOT=/system
 mount -o bind /dev/urandom /dev/random
-umount_all
-mount_all
-mount -o rw,remount -t auto /system
-mount -o rw,remount -t auto /vendor 2>/dev/null
+# Play with partitions only if we have an fstab file available
+if [ -e /etc/fstab ]; then
+  umount_all
+  mount_all
+  mount -o rw,remount -t auto /system
+  mount -o rw,remount -t auto /vendor 2>/dev/null
+else
+  # remount ANDROID_ROOT, sometimes necessary if mounted read-only
+  grep -q "$ANDROID_ROOT.*\sro[\s,]" /proc/mounts && mount -o remount,rw $ANDROID_ROOT
+fi
 ui_print " "
 
 # _____________________________________________________________________________________________________________________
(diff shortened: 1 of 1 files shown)


mfonville:
@osm0sis could you also please have a look at this and confirm it doesn't break anything for regular devices?

osm0sis:
No, every Android device has /etc/fstab in recovery. What crazy use-case is this? Booted installation? GApps should never be installed on a running system partition

JulienBolard:
Unfortunately, emulated devices don't have any recovery, that's why you need to install the package when device is running. (the lack of recovery doesn't help for sure)
This ~solution~ hack has been in use since 2013 in Genymotion.

osm0sis:
Presumably Genymotion is like modern devices and emulates both legacy system, and system-as-root devices?

Simply attempting to remount `$ANDROID_ROOT` rw won't be enough, since `$ANDROID_ROOT` will always be `/system` on a booted (in this case emulated) device but, on a system-as-root device that will not be the mountpoint, `/` will, so the remount will fail.

Your solution will work on old images with a ramdisk and therefore /system as a mountpoint, but is incomplete for newer system-as-root ones, so maybe we can come up with something cleaner there.

osm0sis:
In AK3 I recently added avoidance of the mounts when in `$BOOTMODE` (i.e. a booted device), so we can probably do something similar.

https://github.com/osm0sis/AnyKernel3/commit/273384db045b310faa5916d38df6d7be4729083e

JulienBolard:
>Presumably Genymotion is like modern devices and emulates both legacy system, and system-as-root devices?

We are now using system-as-root in Android 10 yes.

> Your solution will work on old images with a ramdisk and therefore /system as a mountpoint, but is incomplete for newer system-as-root ones, so maybe we can come up with something cleaner there.

You are right, in fact, the remounting of / in rw (and previously /system) is already done by the script that is launching "update-binary" script. We have added these lines in it:
```
SYSTEM_ROOT="/system"
# On Android 10 (API 29) we switched to system-as-root      
if [ "$API_VERSION" -ge 29 ]; then                    
    SYSTEM_ROOT="/"                                   
fi 
```

The remounting that is done in installer.sh is useless for us and will generate an error log anyway. I just kept it here but I think it can be removed.

>You also would want to keep the /vendor remount rw for Genymotion, so GApps changes occur as intended there also, no?

We haven't splitted files into /vendor nor /product or /odm partitions yet for simplicity.

osm0sis:
#815 should work for you then, and still has those remounts just in case. :+1:

JulienBolard:
> In AK3 I recently added avoidance of the mounts when in $BOOTMODE

Really like your solution better :+1:

JulienBolard:
Let's close this one an use #815

PR #815, opened 2020-03-06
Installer OUTFD fixes and emulator support

### LINKED ISSUE #813: Installation on Genymotion devices using drag n drop fails
**Issue description:**

Installation on Genymotion devices using drag n drop fails

**Reproduction Steps:**

1. Get the latest x86/x86_64 archive of Opengapps 
2. Drag n drop this archive on a running Genymotion device (desktop or int the Cloud)
3. Genymotion will detect this as an opengapps archive and ask if you want to continue the installation
4. Click yes
5. Genymotion displays an error in the installation process
6. Device is not accessible anymore using adb shell

**Expected behavior:**

Package should be installed correctly and device should indicate the user it needs to be restarted

**Observed behavior:**

Package installation (installer.sh) returns an error and device is now in an unusable state

(Patch incoming from Genymotion)

osm0sis:
installer: fix OUTFD use when redirected to a log, cleanups:
- append to OUTFD to avoid truncating when redirected to a log file
- use ui_print in backend to simplify error output

installer: support emulator drag-and-drop installs:
- Fixes #813, closes #814

### DIFF
--- a/scripts/inc.installer.sh
+++ b/scripts/inc.installer.sh
@@ -1107,12 +1107,21 @@ ui_print " "
 # For reference, check https://github.com/osm0sis/AnyKernel3/blob/master/META-INF/com/google/android/update-binary
 ui_print "- Mounting partitions"
 set_progress 0.01
+
+ps | grep zygote | grep -v grep >/dev/null && BOOTMODE=true || BOOTMODE=false
+$BOOTMODE || ps -A 2>/dev/null | grep zygote | grep -v grep >/dev/null && BOOTMODE=true
+
 test "$ANDROID_ROOT" || ANDROID_ROOT=/system
-mount -o bind /dev/urandom /dev/random
-umount_all
-mount_all
-mount -o rw,remount -t auto /system
+
+# emulators can only flash booted and may need /system (on legacy images), or / (on system-as-root images), remounted rw
+if ! $BOOTMODE; then
+  mount -o bind /dev/urandom /dev/random
+  umount_all
+  mount_all
+fi
+mount -o rw,remount -t auto /system || mount -o rw,remount -t auto /
 mount -o rw,remount -t auto /vendor 2>/dev/null
+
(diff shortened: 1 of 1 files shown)


scripts/inc.installer.sh
```diff
@@ -1106,12 +1107,21 @@ ui_print " "
 # For reference, check https://github.com/osm0sis/AnyKernel3/blob/master/META-INF/com/google/android/update-binary
 ui_print "- Mounting partitions"
 set_progress 0.01
+
+ps | grep zygote | grep -v grep >/dev/null && BOOTMODE=true || BOOTMODE=false
+$BOOTMODE || ps -A 2>/dev/null | grep zygote | grep -v grep >/dev/null && BOOTMODE=true
```
JulienBolard: Maybe we can use getprop here instead using
[init.svc.zygote]: [running]

osm0sis: This is the established script for $BOOTMODE, established by Magisk, and is known working in all cases, so I'd rather not mess with it.

osm0sis:
@JulienBolard can you confirm an Open GApps zip built with this PR merged works?

JulienBolard:
Current builds on 'master' branch for api 28 and 29 get installed successfully after drag n drop, BUT I have a boot loop when devices rebooted.
I need to investigate why https://pastebin.com/60pnsUbM

mfonville:
@JulienBolard the error is related to a duplicate instance of the package-manager app. Maybe there is both an AOSP/customized/OpenGApps apk that serves that function installed?

JulienBolard:
> @JulienBolard the error is related to a duplicate instance of the package-manager app. Maybe there is both an AOSP/customized/OpenGApps apk that serves that function installed?

It was the opposite: No package Installer was present after my reboot. It seems to come from the .zip I built. 
It works without any problem with march 7 nightly.

PR #821, opened 2020-03-26
SoundPicker: move into PicoGApps, making it removable 

### LINKED ISSUE #767: "Google Sounds" included in Super?
as above. The apps is listed in the "Links to Google Apps" but there's no entry on the default gapps-config.txt

Thanks

Rupurudu:
SoundPicker is not a critical component as all ROMs have a Media storage or a file explorer to pick sounds.

Fixes #767.

Changes:
- Moves SoundPicker from core to pico package
- Makes it configurable with tag "SoundPicker"
- Adds it as an option in the AROMA version.

### DIFF
--- a/scripts/inc.aromadata.sh
+++ b/scripts/inc.aromadata.sh
@@ -181,6 +181,7 @@ form(
       "Sheets",     "<b>Google Sheets</b>",       "",                      "check",
       "Slides",     "<b>Google Slides</b>",       "",                      "check",
       "Search",     "<b>Google Search</b>",       "To Exclude Google Search AND Google Now Launcher/Pixel Launcher <#f00>OR</#> To Include Google Search",                      "check",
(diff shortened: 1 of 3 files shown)


mfonville:
Thanks!

curiousrom:
@mfonville - Could Sounds (com.google.android.soundpicker) be removed from Pico to keep it minimal & be available only on Nano & up?

mfonville:
Technically it could I think, @nezorflame how do you feel about that?

nezorflame:
I agree, the initial decision to do it that way was from @SpasilliumNexus, right now I don't see the need to keep it in Pico.

mfonville:
@curiousrom if you would make a PR for the change, I will merge it in :-)

nezorflame:
See https://github.com/opengapps/opengapps/pull/829

PR #824, opened 2020-03-28
Update launcher_list in installer script

razorloves:
Looks like the launcher_list hasn't been updated in 4 years, so just adding a few newer ones.

Changes:
- Add newest Launcher3 and Trebuchet targets


mfonville:
Thanks!

PR #826, opened 2020-03-28
installer: fix exxit unmounting

aleasto:
* cleaning up $TMP removes busybox and its symlinks, causing
   umount to fail with `umount: not found`

 * This went unnoticed cause TWRP and older Lineage Recovery
   shipped their toolbox in /sbin, which was available at the time
   of unmounting.
   Lineage Recovery 17.1 however, ships its toybox in /system/bin,
   which is unavailable when unmounting because android's system
   is mounted at /system at that moment.

### DIFF
--- a/scripts/inc.installer.sh
+++ b/scripts/inc.installer.sh
@@ -1249,7 +1249,7 @@ exxit() {
     tar -cz -f "$log_folder/open_gapps_debug_logs.tar.gz" logs/*
     cd /
   fi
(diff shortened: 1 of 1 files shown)


npjohnson:
This fixes installation on SAR in Lineage (AOSP 10 based) recovery.

This should've been like this to begin with, as well, I believe.

mfonville:
Thanks!

PR #830, opened 2020-04-20
installer + addon.d-v2 fixes for Lineage 17.1

osm0sis:


osm0sis:
Looks like addon.d might need another fix. Please wait to merge.

I'll ping you. :+1:

scripts/inc.installer.sh
```diff
@@ -2822,14 +2822,25 @@ bkup_header="#!/sbin/sh
 #
 # /system/addon.d/70-gapps.sh
 #
-. /tmp/backuptool.functions
 
-if [ -z \$backuptool_ab ]; then
-  SYS=\$S
-  TMP="/tmp"
+V1_FUNCS=/tmp/backuptool.functions
```
aleasto, line 2831: this will get sed'ed to `/postinstall/tmp/backuptool.functions` by backuptool.sh
but you can trust $backuptool_ab without having to do this

osm0sis: It actually won't because there sed is specifically looking for ". /tmp/backuptool.functions" but since this isn't a complex addon.d script which would require that I'll remove that commit.

osm0sis:
@mfonville, should be good now! :+1:

mfonville:
@osm0sis I assume it has been tested on devices themselves?

osm0sis:
Yep. The Lineage guys found and tested the command, and tested out the changes already. :+1:

PR #835, opened 2020-05-04
Move overlays to /product, add Google Dialer overlay

nezorflame:
`/product` seems to be a better place after all.
Also added `DefaultDialerOverlay` overlay to finally support making Google Dialer the default.
See [this PR](https://gitlab.opengapps.org/opengapps/all/-/merge_requests/11) for the related changes to the overlays.


BealeStBluesBoy:
This would solve my problem I think? On Lineage, since some devices have verity enabled on vendor partition (like mine) installing stock package results in pixel launcher fc because overlay is not being installed.

mfonville:
@nezorflame should this be merged?

nezorflame:
@mfonville yup, tested and confirmed working

PR #840, opened 2020-06-03
installer: Support vendor and product dynamic partitions

razorloves:
This is needed so that the installer script can remove stock/aosp  
items on /product and install overlay apk's to /vendor.


osm0sis:
@mfonville, looks good to me! :+1:

mfonville:
Thanks!

PR #845, opened 2020-06-15
Add ArrowOS apps to removal lists

rapperskull:


mfonville:
Great, thanks @rapperskull

PR #846, opened 2020-06-22
Fix inclusion of gmssetup

razorloves:
The gmssetup apk is being included in api 28 and greater opengapps
zips, but google replaced gmssetup with backuprestore in api 28 and
greater, so fix inclusion logic so that gmssetup is not included in api 28+.


nezorflame:

PR #847, opened 2020-06-22
Add actionsservices to super only on API 28

rapperskull:
Avoids installing actionsservices twice on Android 10 with super package


nezorflame:

PR #848, opened 2020-06-26
Add tvmini variant

webgeek1234:
This splits tv gapps into tvmini and tvstock. Stock has everything it had before, while mini only contains what is required for boot (like the launcher) and what is not available on the play store.


nezorflame:

PR #850, opened 2020-07-04
Add Google Recorder to `full` variant, various fixes

nezorflame:
- Added: Google Recorder to the `full` variant for SDK28 and up
- Fixed: Chrome and Webview are now properly picked for SDK29
- Fixed: set SELinux context for overlays
- Fixed: added overlays to the cleanup list
- Fixed: repack libs inside of the APK only in cases when there are compresed libs inside (fixes the missing signature issue for most of the cases)
- Fixed: zipalign only APKs with v1 signature from jarsigner
- Added: `apksigner` utility requirement for the APK signature checks

### DIFF
--- a/scripts/build_gapps.sh
+++ b/scripts/build_gapps.sh
@@ -66,7 +66,7 @@ ZIPCOMPRESSIONLEVEL="0" # Store only the files in the zip without compressing th
 . "$SCRIPTS/inc.sourceshelper.sh"
 
 # Check tools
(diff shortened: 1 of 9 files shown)


mfonville:
Do we have an update if this has been tested succesfully?

nezorflame:
Unfortunately I still don't have any thorough reports - I did check it myself and it worked (arm64), should work for ARM devices as well since fixes are mostly universal.

nezorflame:
@mfonville got more reports, seems fine. Let's merge this.

PR #852, opened 2020-07-07
Improve discovery of the `$TOP` directory

rhansen:
Rather than assume that `$PWD` is also `$TOP`, discover `$TOP` by `cd`ing to the directory containing the script. This assumes that the script is not executed via symlink, or, if it is, that the symlink is in the same directory as the script itself.

Also eliminate the remaining uses of `realpath`.

### DIFF
--- a/add_sourceapp.sh
+++ b/add_sourceapp.sh
@@ -12,8 +12,7 @@
 #    GNU General Public License for more details.
 #
 
-command -v realpath >/dev/null 2>&1 || { echo "realpath is required but it's not installed, aborting." >&2; exit 1; }
-TOP="$(realpath .)"
+TOP=$(cd "${0%/*}" && pwd -P) || exit 1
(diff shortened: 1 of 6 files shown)


mfonville:
Thanks, does this also work for MacOS?

rhansen:
> Thanks, does this also work for MacOS?

I see no reason why it wouldn't, though I don't have a MacOS machine myself.

mfonville:
@mc1100 didn't you have MacOS? Could you verify/check that it works correctly?

mc1100:
Those particular changes introduced in PR will work on MacOS. But there are still other issues when building on macOS. Namely, some "native" macOS tools used during the build are not compatible with their GNU counterparts.

mfonville:
Thanks! Hopefully we can still fix those other non-compatible things, I'll merge this :-)

PR #854, opened 2020-07-12
Add launcherhack to fix pixellauncher overlay issue

razorloves:
The pixel launcher overlay apk gives exclusive special permissions to
the pixel launcher app for the Recents UI. This causes aosp/stock
launchers to crash. The pixel launcher overlay is only installed
in api28 and newer, so this change adds launcher to the removal list
in api28+ to fix this problem.

### DIFF
--- a/scripts/inc.compatibility.sh
+++ b/scripts/inc.compatibility.sh
@@ -284,6 +284,17 @@ kitkatpathshack(){
   fi
 }
 
+launcherhack(){
+  if [ "$API" -ge "28" ]; then
+    cat <<'EOFILE'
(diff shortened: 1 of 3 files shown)


mfonville:
Thanks! Have you tested the change on your device?

osm0sis:
@razorloves, is this still needed? And can you confirm it's tested and working for @mfonville?

nezorflame (approved):
Sorry, very late on this, LGTM

razorloves:
yep, still needed and tested/working.

mfonville:
Thanks!

PR #856, opened 2020-07-14
Fix installer template to respect preodex config.

### LINKED ISSUE #855: Installer template will never `dex2oat` anything, even when asked.
**Only issues related to the code of The Open GApps Project scripts themselves**

If you have found an issue with a pre-built OpenGApps.org package please use the XDA-Developers [forum thread](http://forum.xda-developers.com/android/software/Open-GApps-t3098071)

**Checklist**

- [X] The issue is not answered by reading the [FAQ](https://github.com/opengapps/opengapps/wiki/FAQ)
- [X] The issue is part of The Open GApps Project scope *(and not of the pre-built OpenGApps.org packages scope, for support on those see [XDA Forum](http://forum.xda-developers.com/android/software/Open-GApps-t3098071))*

*If Open GApps compilation fails:*
- [X] The issue is not a [duplicate](https://github.com/opengapps/opengapps/issues?q=is%3Aissue)

*If Open GApps installation fails:*
- [X] Is the issue reproducible using the latest version of your recovery?
- [X] Is the issue reproducible with (the latest) [TWRP](https://twrp.me) recovery? *(other recoveries are only best-effort)*

*If receiving force closures after Open GApps installation:*
- [X] The issue is **NOT** an error within one of the bundled Google applications itself *(verify by e.g. installing the application from the Play Store instead)*
- [X] The issue is reproducible on clean (=wiped ``/data/`` partition) installation
- [X] The affected ROM includes the necessary patches and functionality as noted for [Android versions](https://github.com/opengapps/opengapps/wiki#notes-for-android-versions) and [Platforms](https://github.com/opengapps/opengapps/wiki#notes-for-platforms)
- [X] No non-AOSP hacks like privacy guards' permission restrictions etcetera are applied

**Issue description:**

If a configuration is presented which should trigger `preodex`, the installer script will fail to ever run `dex2oat` due to incorrect logic when checking for a dex file.

**Reproduction Steps:**

1. A `Preodex` to a config file
2. Run an update, watch it never run `dex2oat`

**Expected behavior:**

Run `dex2oat` when the config deems it necessary.

**Observed behavior:**

No `dex2oat` is run.

-----

This is because the `check_manifest` will return `1` for anything which containers a `classes.dex` per https://github.com/opengapps/opengapps/blob/master/scripts/templates/installer.sh#L1118. However when this function is called, it is looking for the negative of this result, as seen https://github.com/opengapps/opengapps/blob/master/scripts/templates/installer.sh#L1332. This means it will only ever try to run `dex2oat` on things which _do not_ contain a `classes.dex` file, which is the incorrect logic.

strazzere:
Fixes #855.

Changes:
- Ensure the installer template correctly processes the `checkmanifest` function result

### DIFF
--- a/scripts/templates/installer.sh
+++ b/scripts/templates/installer.sh
@@ -1329,7 +1329,7 @@ install_extracted() {
     true*)
       installedapkpaths="$(find "$TMP/$1/" -name "*.apk" -type f | cut -d/ -f5-)"
       for installedapkpath in $installedapkpaths; do  # TODO fix spaces-handling
-        if ! checkmanifest "/system/$installedapkpath" "classes.dex"; then
+        if checkmanifest "/system/$installedapkpath" "classes.dex"; then
           ui_print "- pre-ODEX-ing $gapp_name"
           log "pre-ODEX-ing" "$gapp_name"
           odexapk "/system/$installedapkpath"


vpnsade:
#855

mfonville:
@osm0sis can you also have a look please?

strazzere:
As an update to this - the preodex is "broken" as stated somewhere in the code (failing to find the reference). So I ended up _not_ using this config value. However, this fix to ensure it's respected is still "valid".

I'm unsure if the preodex function is actually used anywhere, or if it is left over from something previous though?

It may be worth refactoring and removing it all together, which I will gladly due, however I didn't want to remove it and say "it didn't work on my device" in case it's used elsewhere and works.

osm0sis:
@mfonville, I'm not familiar with the preodex stuff, I seem to recall that was your baby. :wink:

mfonville:
It was indeed an experimental feature by me. But I think we can safely drop the code as it is probably to outdated anyhow

strazzere:
@mfonville / @osm0sis if neither of you have an objection - I can just close this issue and pull request, then open a new one where these preodex code paths are fully removed.

Thoughts?

osm0sis:
Yep go for it

mfonville:
Yes, that's a good idea!

PR #865, opened 2020-08-10
installer: Fix ch_con function calls

razorloves:
The ch_con function was renamed to ch_con_system in a previous commit
but all calls were not renamed, so this change fixes that.
Addresses following errors during zip install:
/tmp/installer.sh: line 1454: ch_con: not found
/tmp/installer.sh: line 2958: ch_con: not found
/tmp/installer.sh: line 2961: ch_con: not found


nezorflame:
Damn, missed that in my PR. Thanks @razorloves!

mfonville:
Thanks!

PR #868, opened 2020-08-16
installer: mount /product for non dynamic devices as well

merothh:
* As we dont mount /product for devices that are not dynamic
  and have a separate partition, on such devices AOSP counterpart
  apps in /product end up not being removed, among possibly other
  similar issues as well.

* So mount it.

  Thanks to @nikhilmenghani for the tip

### DIFF
--- a/scripts/templates/installer.sh
+++ b/scripts/templates/installer.sh
@@ -939,6 +939,7 @@ mount_all() {
   fi
   (mount /cache
   mount -o ro -t auto /persist
+  mount -o ro -t auto /product
   mount -o ro -t auto /vendor) 2>/dev/null
   setup_mountpoint $ANDROID_ROOT
   if ! is_mounted $ANDROID_ROOT; then


merothh:
Verified on Redmi Note 7 Pro (Non Dynamic, but uses a separate product partition in custom ROMs)

merothh:
I'm also doing a follow up change for addon.d

nezorflame:
Hi, thanks for your PR!
@osm0sis can you take a look as well?

osm0sis:
Makes sense, I guess!

mfonville:
So I wait for a merge until addon.d is also updated?

osm0sis:
addon.d will be tricky, I'm not even sure if/how we handle /vendor there currently

merothh:
> So I wait for a merge until addon.d is also updated?

Yes, that would be ideal.

> addon.d will be tricky, I'm not even sure if/how we handle /vendor there currently

Indeed. I didnt see anything to handle /vendor. 

https://github.com/merothh/opengapps/commit/afa50faffaaeda32bedcbf91c718dff8118f4709

This was what i came up with then, it was working on non dynamic. But the dynamic one was screwing up and aborting the install when i dirty flashed. Not sure why, but i guess it was because the product partition isn't mapped when it tries to mount at pre-backup ? 

If we do manage to get it working somehow, might as well extend to it mount vendor as well (if needed).

osm0sis:
So probably better to just merge this and have addon.d on a to-do list for later.. I'm thinking if we can get things mounted right then removals should be able to happen via the normal addon.d functions since it's all symlinked into /system.

merothh:
> So probably better to just merge this and have addon.d on a to-do list for later.. I'm thinking if we can get things mounted right then removals should be able to happen via the normal addon.d functions since it's all symlinked into /system.
> 
> But on dynamic, most importantly on A/B devices with addon.d-v2, those extra mounts will need to be into /postinstall, and are not standard to do, so it'll be pretty tricky.

Alright then.

mfonville:
Alright, thanks!

nezorflame:
Let's track the addon.d issue [here](https://github.com/opengapps/opengapps/issues/872)

PR #869, opened 2020-08-16
installer: simplify ch_con* usage

osm0sis:
- No need for extra functions
- No need to echo | grep, when we can just case


osm0sis:
@mfonville, this one too, please. :+1:

Issue #872, opened 2020-08-20
Proper addon.d support for the /vendor and /product partitions for A/B devices


nezorflame:
Thanks to @merothh users who have custom /product partition are now able to properly remove stock apps during the installation process: https://github.com/opengapps/opengapps/pull/868

The addon.d script though requires more work and stil needs to be fixed. Let's track this issue here.

osm0sis:
/vendor will also need some consideration.. for addon.d (v1) they can be mounted to /vendor and /product, but for addon.d-v2 it'll need to be /postinstall/vendor and /postinstall/product, but only if e.g. /postinstall/product is a directory and /postinstall/system/product is a symlink.

merothh:
https://github.com/merothh/opengapps/commit/e2266611b081142b8fef0354fccf52f06897fc1c

osm0sis:
Nice work!!

A couple minor notes:

1) Please make the mount_product function mount_generic so that we can just supply the word product or vendor and it'll do the setup.
2) It will need to mount to /postinstall/product or /postinstall/vendor for dynamic and/or addon.d-v2.

If you could fix those up and submit as PR I think we're good! 🙂

merothh:
As for:

1. That's what I wanted to do. Just wanted to see if this initial approach was feasible. I'll do it like so.

2. I thought I'd leave that since I don't have an A/B device to test on. Hence I just left checks to skip it on A/B. I'll see to that as well then.

merothh:
> Nice work!!
> 
> A couple minor notes:
> 
> 1. Please make the mount_product function mount_generic so that we can just supply the word product or vendor and it'll do the setup.
> 2. It will need to mount to /postinstall/product or /postinstall/vendor for dynamic and/or addon.d-v2.
> 
> If you could fix those up and submit as PR I think we're good! 🙂

As for:

1: That's what I wanted to do. Just wanted to see if this initial approach was feasible. I'll do it as you suggested.

2: I thought I'd leave that since I don't have an A/B device to test on. Hence I just left checks to skip it on A/B. I'll see to that as well then.

osm0sis:
Hmm there might still be a complication with A/B.. I think addon.d using $SYS or /postinstall/system as a base will cause issues, since the /system/vendor and /system/product symlinks aren't relative.. so would point to actual /vendor and /product, which would be kinda haphazard to do in a global mount namespace on a booted device.. 😕

merothh:
[...]

PR #879, opened 2020-09-04
addon: handle mounting of {product,vendor} on A only devices

merothh:
* On devices with a separate product partition, addon scripts were
  not able to remove AOSP counterpart apps among possibly other
  similar issues as well

* So fix it by mounting product before we get started

* While we're at it, mount vendor too since its mounted in the
  main installer script

* Thanks to @NikhilMenghani for his input

Fixes https://github.com/opengapps/opengapps/issues/872 for A only devices.

### DIFF
--- a/scripts/bkup_tail.sh
+++ b/scripts/bkup_tail.sh
@@ -1,6 +1,25 @@
 EOF
 }
 
+mount_generic() {
+  local active_slot=$(getprop ro.boot.slot_suffix)
+  local partitions="$*"
+  if [ -z "$active_slot" ]; then
+    # We're on an A only device
+    local partition
+    for partition in $partitions; do
+      if [ "$(getprop ro.boot.dynamic_partitions)" = "true" ]; then
+        mount -o ro -t auto /dev/block/mapper/"$partition" /"$partition" 2> /dev/null
+        blockdev --setrw /dev/block/mapper/"$partition" 2> /dev/null
+        mount -o rw,remount -t auto /dev/block/mapper/"$partition" /"$partition" 2> /dev/null
+      else
+        mount -o ro -t auto /"$partition" 2> /dev/null
+        mount -o rw,remount -t auto /"$partition" 2> /dev/null
+      fi
+    done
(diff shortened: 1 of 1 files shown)


nezorflame (approved):
LGTM

nezorflame:
@osm0sis @mfonville please take a look as well

osm0sis:
Only note I have would be to make sure the variables used are local only, otherwise looks good!

mfonville:
@merothh thanks for your work!
Did you test this on devices? :-)

merothh:
> @merothh thanks for your work!
> Did you test this on devices? :-)

Indeed i had tested the initial commit i PR-ed on **A only (Both Dynamic / Non Dynamic)**. As noted in  the last [comment](https://github.com/opengapps/opengapps/issues/872#issuecomment-682970129) i made on the [issue](https://github.com/opengapps/opengapps/issues/872)

While i was adding the changes osm0sis mentioned [above](https://github.com/opengapps/opengapps/pull/879#issuecomment-687134674), i realized how arrays weren't exactly POSIX, so i've avoided using arrays as i did initially.

The latest commit has been again tested on **A only (Both Dynamic / Non Dynamic)**, and after you guys have one last look, it seems go to go.

merothh:
Diff between initial and latest commit

osm0sis:
I think it looks like it'll work fine. My only thought is there's probably no reason to mount twice, just mount rw the first time for each.

Also, minor caveat, I seem to recall there being some devices that report they're slot _a even though that's only only slot on the device, but it's hard to deal with broken device setups like that.

mfonville:
Thanks for your comments @osm0sis 
I think first mounting read-only and then trying to remount with rw is the same (good) behavior was we have in the installer script, because of some edge-cases (like emulators).

I will now merge this in, thanks @merothh !

If we can think of a way to detect the broken slot_a setups that would be very welcome in a new PR :-)

osm0sis:
Ah yes, I always forget about the emulators

Issue #889, opened 2020-10-30
Only tvstock/tvmini builds available for Android 8.0(missing pico+others)


ubergeek42:
I'm only seeing tvstock/tvmini builds for 8.0(I need pico on x86, but they seem to be missing for arm as well). The last build of x86/pico I could find was from June 30.

I saw #862 , but it was closed as resolved.

nezorflame:
Thanks for the report! It was due to a bug in the build script, should be fixed when the PR mentioned above is merged.

nezorflame:
Since this has been fixed, I'm closing this.

PR #893, opened 2020-11-26
Fix GMS issue for SDK27 and lower

nezorflame:
Fixes missing 8.0 builds for everything non-TV related - there are no SDK26-specific GMS packages, only SDK23 and SDK27+. This caused the builds for SDK26 to fail since July (!).
So, we use SDK23 as the minAPI one.

### DIFF
--- a/scripts/inc.compatibility.sh
+++ b/scripts/inc.compatibility.sh
@@ -303,10 +303,8 @@ minapihack(){
         useminapi="24"
       fi;;
     com.google.android.gms)
-      if [ "$API" -ge "28" ]; then
-        useminapi="28"
-      elif [ "$API" -ge "26" ]; then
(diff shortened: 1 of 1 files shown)


nezorflame:
Closes #889

nezorflame:
Fixes #889

osm0sis:
Is the more of a workaround? Not familiar with this part of things, so I'll hold off and let @mfonville check it out. 👍

nezorflame:
Well, technically it's a workaround, but we use it for multiple packages. Now it's more like a feature :)

mfonville:
Thanks! I will merge it :-) 

@NicholasBuse had you notice the change in GMS sdk versions, and do we need a clean-up somewhere?

PR #902, opened 2021-01-01
installer: Fix support for Android 11 Recovery with system as root

AdrianDC:
* Currently, if we install OpenGApps from an Android 11 recovery,
   like Lineage Recovery (from LineageOS 18.1 releases), the
   installation will fail due to 0 bytes of free space available,
   and the /system/bin/ recovery binaries from system as root
   will be lost as the system partition is mounted to /system

 * To avoid a system as root breakage, acquire the system source
   and mount it to /system_root instead, and if mounted, ensure
   it is a system as root partition, otherwise we fallback
   to the ANDROID_ROOT mounting point as it used to be

### DIFF
--- a/scripts/templates/installer.sh
+++ b/scripts/templates/installer.sh
@@ -932,6 +932,38 @@ umount_apex() {
   unset ANDROID_RUNTIME_ROOT ANDROID_TZDATA_ROOT BOOTCLASSPATH
 }
 
+mount_system_source() {
+  local system_source
+  system_source=$(grep ' /system ' /etc/fstab | tail -n1 | cut -d' ' -f1)
+  if [ -z "${system_source}" ]; then
+    system_source=$(grep ' /system_root ' /etc/fstab | tail -n1 | cut -d' ' -f1)
+  fi
+  if [ -z "${system_source}" ]; then
+    system_source=$(grep ' / ' /etc/fstab | tail -n1 | cut -d' ' -f1)
+  fi
+  echo "${system_source}"
+}
+
+mount_android_root() {
+  local system_source
+  system_source=$(mount_system_source)
+  if [ ! -z "${system_source}" ]; then
+    setup_mountpoint /system_root
+    if ! is_mounted /system_root; then
+      mount -o ro -t auto "${system_source}" /system_root 2>/dev/null
+    fi
+    if is_mounted /system_root && [ ! -f /system_root/system/build.prop ] &&
+        ! is_mounted $ANDROID_ROOT && [ ! "$ANDROID_ROOT" = '/system_root' ]; then
+      mount --move /system_root "${ANDROID_ROOT}"
+    fi
+  else
+    setup_mountpoint $ANDROID_ROOT
+    if ! is_mounted $ANDROID_ROOT; then
+      mount -o ro -t auto $ANDROID_ROOT 2>/dev/null
+    fi
+  fi
+}
+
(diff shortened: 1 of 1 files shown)


AdrianDC:
osm0sis

althafvly:
Android -11 Lineage recoveries needs this support

@nezorflame can you check opengapps gitlab. I have registered there, so i can pull request arm changes for eleven

nezorflame:
Thanks for your contribution!
So far seems OK, but need a bit more insight here.
@mfonville @osm0sis please take a look as well.

nezorflame:
I'll also need you to test a build with these changes on two devices - with the described setup and without it (maybe an A-only device or some legacy one).

mfonville:
Thanks! This definitely needs some testing and review by @osm0sis :-)

AdrianDC:
I plan to do some more minor changes, will have time for it in the coming days. Thanks.

osm0sis:
Hmm I'm currently working through this for AK3 on LOS 18.1 Recovery and I see on dynamic partition devices that /system_root, /system and / are no longer in /etc/fstab so I don't think this case is covered yet.

osm0sis:
Alright, I've been informed that /dev/block/mapper gets setup when a zip is flashed, so that's all good, but I still don't see how the fstab fallback could work. I've just finished my fixes for AK3 and they work on LOS 18.1 Recovery and TWRP_10 in my testing, so perhaps Open GApps could adapt those instead?

Also worth mentioning the concern above about leaving a working shell isn't really important since Open GApps bundles its own busybox which is used instead, similar to Magisk and AK3, but I have attempted to ensure there's always a working underlying shell anyway by reorganizing when the apex mounts happen. :+1:

https://github.com/osm0sis/AnyKernel3/commits/master

mfonville:
@osm0sis I think following your fixes for AK3 sounds good. Will you make a new PR for that?

mfonville:
@AdrianDC could you check @osm0sis's PR?

AdrianDC:
I'll try to have a look at these within the week, but not sure when yet.

@osm0sis 's PR looks OK to me, however I think it doesn't cover the /etc/fstab extraction part of this PR,
meaning the fallback implementation done in OpenGApps to use the ".../system" partition
does not support non-standard situations (for example, "System" here).

osm0sis:
What do you mean, it's /dev/block/<something>/System on some devices?

AdrianDC:
Yes, the same thing applies to bootimage partitions, for example this is what SuperSU (or now Magisk) used to do:
```
for PARTITION in kern-a KERN-A Kernel kernel KERNEL boot BOOT lnx LNX; do
  BOOTIMAGE_PATH=/dev/block/by-name/${PARTITION};
```

You're correct, the `mount /system` then `mount --move /system /system_root` is supposed to work, however,
the `mount /system` itself was masking the /system/bin/mount binaries, thus breaking most following shell commands
and leaving the device in a rather bad situation were only launched things still work.

I'd have to test again when I can, but I'd expect the same overlap behaviour.

osm0sis:
The `mount /system` is actually intentional, and doesn't break anything for Open GApps since all is within Open GApps' included busybox ash sub-shell with applets in the $PATH, that said, I've adjusted when /apex gets mounted now, so that even when /system is mounted the actual /system/bin binaries now actually run, so there shouldn't be a gap of underlying shell breakage.

PR #903, opened 2021-01-01
installer: Wrap the OpenGApps logo to support Android Recovery

AdrianDC:
* Without leading characters, the 'ui_print' calls strip every
   leading spaces, and wrapping with \', \" or \ prints them


AdrianDC:
Different solutions welcomed, but this is a good start IMO.

AdrianDC:
@osm0sis

nezorflame (approved):

nezorflame:
LGTM.
WDYT @mfonville @osm0sis ?

mfonville:
What does Android Recovery do with ASCII-code 255, the non-breaking space? Does it print it or does it also strip it?

osm0sis:
Oof, couldn't Lineage just make Android Recovery display them like TWRP always has?

osm0sis:
> What does Android Recovery do with ASCII-code 255, the non-breaking space? Does it print it or does it also strip it?

Just tested, it shows as ?, unfortunately.

mfonville:
Would it make sense to file a bug report with Lineage to ask them to display like TWRP and legacy CWM and not stripping output in their recovery? As @osm0sis mentioned, why couldn't they just stick to the same behavior that was already many years in place and wasn't broken to begin with?

osm0sis:
Further, I just tried the following in a zip:
```
  i=0;
  while [ $i -lt 400 ]; do
    ui_print "$(printf "%b" "\0$i")     test $i"
    i=$((i +1));
  done;
```

And unfortunately there aren't any viable blank space candidates to prepend to an existing line and keep Lineage from stripping the leading space.

osm0sis:
I'm going to set this back to #wontfix unless anyone has anything else to add?

PR #885, opened 2020-09-30
Android R bringup

nezorflame:
Initial bringup, probably more changes will follow soon.

Corresponding PR's for the APKs:
[all](https://gitlab.opengapps.org/opengapps/all/-/merge_requests/37)
[arm](https://gitlab.opengapps.org/opengapps/arm/-/merge_requests/17)
[arm64](https://gitlab.opengapps.org/opengapps/arm64/-/merge_requests/22)
[x86](https://gitlab.opengapps.org/opengapps/x86/-/merge_requests/9)
[x86_64](https://gitlab.opengapps.org/opengapps/x86_64/-/merge_requests/4)

PR #905, opened 2021-01-14
installer: mounting script updates and function logic fixes

osm0sis:
- fix /system/bin/* execution on recoveries using LD_CONFIG_FILE
- add additional BOOTCLASSPATH entry for /apex mounts
- try to preserve a working /etc/fstab on AOSP Q-based recovery and newer
- fix sloppy/unpredictable `cmd && this || that` statements by being a bit more POSIX to avoid any potential issues when stdout/err are redirected
- do apex mounts before /system bind to always ensure a working underlying shell environment
- improve support for mounting dynamic/logical partitions from /dev/block/mapper
- don't rely on props if possible: determine dynamic/logical partitions from /dev/block/mapper presence, and detect slots more robustly using props and cmdline
- clean up all use of test

Closes #902

### DIFF
--- a/scripts/templates/installer.sh
+++ b/scripts/templates/installer.sh
@@ -860,9 +860,7 @@ nogooglewebview_removal_msg="NOTE: The Stock/AOSP WebView is not available on yo
 
 # _____________________________________________________________________________________________________________________
 #                                                  Pre-define Helper Functions
-get_file_prop() {
-  grep -m1 "^$2=" "$1" | cut -d= -f2
-}
+get_file_prop() { grep -m1 "^$2=" "$1" | cut -d= -f2-; }
 
 set_progress() { echo "set_progress $1" >> $OUTFD; }
 
@@ -871,25 +869,37 @@ ui_print() {
     ui_print" >> $OUTFD
 }
 
+find_slot() {
+  local slot=$(getprop ro.boot.slot_suffix 2>/dev/null)
+  [ "$slot" ] || slot=$(grep -o 'androidboot.slot_suffix=.*$' /proc/cmdline | cut -d\  -f1 | cut -d= -f2)
+  if [ ! "$slot" ]; then
+    slot=$(getprop ro.boot.slot 2>/dev/null)
+    [ "$slot" ] || slot=$(grep -o 'androidboot.slot=.*$' /proc/cmdline | cut -d\  -f1 | cut -d= -f2)
+    [ "$slot" ] && slot=_$slot
+  fi
+  [ "$slot" ] && echo "$slot"
+}
+
 setup_mountpoint() {
-  test -L $1 && mv -f $1 ${1}_link
+  [ -L $1 ] && mv -f $1 ${1}_link
   if [ ! -d $1 ]; then
     rm -f $1
-    mkdir $1
+    mkdir -p $1
   fi
 }
 
 is_mounted() { mount | grep -q " $1 "; }
 
 mount_apex() {
-  test -d /system/apex || return 1
+  [ -d /system_root/system/apex ] || return 1
   local apex dest loop minorx num
   setup_mountpoint /apex
-  test -e /dev/block/loop1 && minorx=$(ls -l /dev/block/loop1 | awk '{ print $6 }') || minorx=1
+  minorx=1
+  [ -e /dev/block/loop1 ] && minorx=$(ls -l /dev/block/loop1 | awk '{ print $6 }')
   num=0
-  for apex in /system/apex/*; do
+  for apex in /system_root/system/apex/*; do
     dest=/apex/$(basename $apex .apex)
-    test "$dest" = /apex/com.android.runtime.release && dest=/apex/com.android.runtime
+    [ "$dest" = /apex/com.android.runtime.release ] && dest=/apex/com.android.runtime
     mkdir -p $dest
     case $apex in
       *.apex)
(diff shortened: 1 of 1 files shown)


scripts/templates/installer.sh
```diff
@@ -956,45 +967,60 @@ mount_all() {
...
+  umount /vendor
+  umount -l /vendor
+  umount /mnt/vendor
+  umount -l /mnt/vendor
+  umount /product
+  umount -l /product
+  umount /mnt/product
+  umount -l /mnt/product
+  umount /persist
+  umount -l /persist
```
mfonville, line 1016: why not umount `/mnt/persist` too?

And these entries after unmounting after apex can still be a for-loop, right?

scripts/templates/installer.sh
```diff
@@ -1171,10 +1201,14 @@ exxit() {
...
     cp -f "$rec_tmp_log" "$TMP/logs/Recovery_tmp.log"
-    curLD="$LD_LIBRARY_PATH"
-    unset LD_LIBRARY_PATH
+    OLD_LD_PATH=$LD_LIBRARY_PATH
+    OLD_LD_PRE=$LD_PRELOAD
+    OLD_LD_CFG=$LD_CONFIG_FILE
+    unset LD_LIBRARY_PATH LD_PRELOAD LD_CONFIG_FILE
     logcat -d -f "$TMP/logs/logcat"
-    export LD_LIBRARY_PATH="$curLD"
+    test "$OLD_LD_PATH" && export LD_LIBRARY_PATH=$OLD_LD_PATH
```
mfonville, line 1209: I just want to mention that  [ expression ] is the same as `test` in Unix.

I assume that quite some code is extracted from your other project, so I am not going to complain about any code-style, because that would only increase any maintenance burden. Best is code to keep shared code as alike as possible for maintainability.

mfonville (commented):
Thanks @osm0sis !

Of course the big question: did you test this on your devices?

@nezorflame could you maybe make a test-spin for users to try-out?

nezorflame:
Yup, this is going in the next test build for sure.

scripts/templates/installer.sh
nezorflame: My idea would be to use shellformat, aka shfmt

scripts/templates/installer.sh
osm0sis: Didn't know /mnt/persist was a thing, and arguably you could just mount it twice for your purposes (.gapps-config).

Also, yes, they could be a for-loop, but I chose to leave them as individual commands so it's closer to my version in AnyKernel3 for the same upkeep reasons you bring up below. 👍

scripts/templates/installer.sh
osm0sis: Yep, I'm aware those built-ins are aliases, but you're right, the idea was to leave it matching the source for upkeep reasons, just as it was on the previous merges for it. I'll consider updating this style tic of mine in the AK3 source in the future so we can all be happy perhaps. 😛

As for shfmt, there is an old Issue, #48, from @mfonville that suggested using ShellCheck linter, if someone was finally wants to pull the trigger on that, which would be cool for sure. 🙂

osm0sis:
> Thanks @osm0sis !
> 
> Of course the big question: did you test this on your devices?

All my usual work testing with the near identical AnyKernel3 scripts. Verified by myself and eng.stk on TWRP_10, and Lineage 18.1 Recovery.

Main differences are with Open GApps setting up and tearing down the flash environment being a bit different (specifically https://github.com/osm0sis/AnyKernel3/blob/master/META-INF/com/google/android/update-binary#L172-#L226). Could by worth considering bringing that over from mine at some point as well to further decrease the maintenance burden here.

scripts/templates/installer.sh
mfonville: I don't know if it is a thing. But it just occurred it is the only one that has no /mnt entry

Keeping it close to your AK3 is good :-)

mfonville:
OK, we definitely need some physical testing next to that the same code works with AK3. Let's wait for Nezorflame's spin and some users testing that before merging it :-)

nezorflame:
I'll leave the links to the test builds here once they're ready.

scripts/templates/installer.sh
osm0sis: Just double-checked on my OnePlus 8 Pro, and on Lineage Recovery the fstab has `
/dev/block/bootdevice/by-name/persist /mnt/vendor/persist` and it's not mounted by default, so I think it should be find to leave as-is, since that won't likely ever get mounted.

osm0sis:
Alrighty, force-pushed to kill "test" with fire, and fixed some improper use of -z that I noticed. :+1:

scripts/templates/installer.sh
osm0sis: Updated AK3 with the additional Open GApps partitions, and added the for-loop. Should be good here now.

scripts/templates/installer.sh
osm0sis: Updated AK3 and the PR to remove all use of test, so this should be good now.

mfonville:
@nezorflame where are we with testing? :-)

mfonville (approved):

nezorflame:
@mfonville @osm0sis unfortunately my local setup is a bit fucked ATM, took all weekend to restore (and not even finished yet).
I hope to bring the new builds tomorrow.

nezorflame:
https://github.com/opengapps/opengapps/pull/885#issuecomment-770253200

mfonville:
@nezorflame Did we get any feedback on the test builds?

nezorflame:
@mfonville no negative feedback regarding the changes, so it should be working fine :)

mfonville:
Alright, time to merge :-)

PR #924, opened 2021-07-12
Move gearheadstub to pico

nezorflame:
Since we've got a lot of requests to add Android Auto stub package to the `pico` as well, here it is.
Any user will be able to exclude it with the `gearheadstub` keyword in gapps-config file.


nezorflame:
@mfonville not sure about the submodule changes, might need to recheck those. I think A11 PR had those changed as well.

PR #931, opened 2021-08-17
Update partitions used by installer and backup scripts

z3DD3r:
Changes:
- Added support for /system_ext partition (mount and search apps for deletion)

### DIFF
--- a/scripts/bkup_tail.sh
+++ b/scripts/bkup_tail.sh
@@ -51,7 +51,7 @@ case "$1" in
       backup_file "$S"/"$FILE"
     done
 
-    umount /product /vendor 2> /dev/null
+    umount /system_ext /product /vendor 2> /dev/null
   ;;
   restore)
     list_files | while read -r FILE REPLACEMENT; do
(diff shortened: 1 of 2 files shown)


nezorflame (approved):
LGTM

osm0sis:
Isn't /persist still potentially used for .gapps-config?

z3DD3r:
> Isn't /persist still potentially needed for .gapps-config? It was being used since there was no other option on some devices (no /data decryption, no /cache) and .gapps-config is tiny. Problems only arose when TWRP was completely filling /persist with logs.

The idea behind removing /persist is that no one should touch it if they don't really need to. In my experience, there have been many problems after users accidentally deleted data from this partition. I understand that this is not an opengapps problem. But if we will not support this partition, then no one will use it and will not break anything...

Your point about /data decryption and /cache is valid! But how user will be able install opengapps in such case? The only way i can imagine is otg. But in case in of otg, gapps-config can be placed near the flashable zip...

Anyway, change for /persist can be easily dropped from this PR. Let know what do you think about it and how i should proceed with this change.

osm0sis:
I think the intention was it's a location to store the .gapps-config on-device, but the zip would need to be sideloaded, so yes, I think it should still be supported. Magisk supports it for sepolicy as well for similar reasons. As far as I know the only issues were with TWRP using it for logs and filling it up.

z3DD3r:
@osm0sis,
Got it! I have removed change for /persist partition

osm0sis (approved):

mfonville (approved):

mfonville:
Thanks!

PR #935, opened 2021-09-07
Remove the googlepixelconfig package

razorloves:
Each of the files in this package are device specific so they  
shouldn't be getting installed to all pixel devices.

### DIFF
--- a/scripts/inc.buildtarget.sh
+++ b/scripts/inc.buildtarget.sh
@@ -20,7 +20,6 @@ googlecontactssync
 googlefeedback
 googleonetimeinitializer
 googlepartnersetup
-googlepixelconfig
 gmscore
 gsfcore
 vending"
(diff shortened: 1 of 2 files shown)


osm0sis:
Last time I looked at the latest Pixel system dumps they contained the xmls from all previous ones. 🤔

razorloves:
Yeah, i didn't say that accurately.  sorry.
Just meant that all the xml's shouldn't be getting installed to all pixels.
pixel 1 only has 2016 xml's.
pixel 2 only has 2017 xml's.
pixel 3 only has 2017 and 2018 xml's
pixel 3a has 2017, 2018, and 2019 mid-year xml's.
and so on...
So, in my opinion, it should just be included in the rom and removed from here.
Otherwise, the conditional statement and xml's in opengapps would have to be added/updated/reworked every time google releases a new device.

osm0sis:
Seems like the "experience" ones carry on through the newer ones and the exclusive ones are single-device. Anyway, I guess that's more of an organizational thing @mfonville or @nezorflame will need to comment on.

razorloves:
One of the things that google uses these files for is settings the cutoff date for the unlimited free photo storage at original quality.
So opengapps on some pixels is currently causing the cutoff date to be wrong.
https://support.google.com/photos/answer/6220791?co=GENIE.Platform%3DAndroid&oco=1

osm0sis:
Yep that's in the exclusive ones.

mfonville:
I am not sure if ROMs can bundle the "exclusive" XML. We probably have to step up our game here to make them device-specific. (so yes, that means every time Google releases a new device, we should add it)

And keep the "experience" in place.

razorloves:
> I am not sure if ROMs can bundle the "exclusive" XML.

i don't see why not.  lineageos has been including it, and the other xml's, in all the pixel device for many years.  
most roms probably already do include it.

razorloves:
These xml's were not included in gapps for years.  Until last year when i noticed there was a googlepixelconfig package but it wasn't being installed.  So I did a [PR](https://github.com/opengapps/opengapps/pull/805) to add it.  Which, in hindsight, was poorly implemented.  So this is basically just a revert of that change of mine.  

Also, i just checked numerous other pixel custom roms and couldn't even find one that doesn't already include these xml's.

IMHO, this can be safely merged now so that it will stop causing the main issue mentioned above, and possibly others.  If someone in the future wants to go above and beyond and re-introduce these xml, they are free to of course.  But I feel it would just quickly become neglected and outdated again, like it is now and has in the past.

PR #942, opened 2021-11-16
[WIP] Reverse ABI check to fix wrong arch issue

nezorflame:
Fixes the "wrong architecture" issue for some ROMs.

Thanks to @nikgapps and @ipdev99 for noticing this.

### DIFF
--- a/scripts/templates/installer.sh
+++ b/scripts/templates/installer.sh
@@ -1663,10 +1663,10 @@ if [ ! "$rom_build_sdk" = "$req_android_sdk" ]; then
 fi
 
 # Check to make certain that user device matches the architecture
-device_architecture="$(get_prop "ro.product.cpu.abilist")"
+device_architecture="$(get_prop "ro.product.cpu.abi")"
 # If the recommended field is empty, fall back to the deprecated one
 if [ -z "$device_architecture" ]; then
-  device_architecture="$(get_prop "ro.product.cpu.abi")"
+  device_architecture="$(get_prop "ro.product.cpu.abilist")"
 fi
 
 case "$device_architecture" in


osm0sis:
Oh, when did that change?. I've always checked /system/build.prop (for e.g. 64 bit devices running 32 bit ROMs) and then default.prop, like this:

https://github.com/Magisk-Modules-Repo/busybox-ndk/blob/master/customize.sh#L314-L334

Magisk does similarly:
https://github.com/topjohnwu/Magisk/blob/master/scripts/util_functions.sh#L30-L46
https://github.com/topjohnwu/Magisk/blob/master/scripts/util_functions.sh#L522-L543

mfonville:
Thanks for noticing and figuring this out. But it is weird, `abilist` is the recommended field. `abi` and `abi2` should be obsolete by now :-/

On which ROMs does this happen, do we have some examples of the prop-files?

I am a bit hesitant about this change, because I fear we could break other ROMs (and e.g. emulators).

If possible I'd prefer a more surgical method to detect these broken ROM prop files and deal with it properly.

nezorflame:
aight, marking this as WIP then

nezorflame:
From what I've seen from @osm0sis links, the only prop checked is the `ro.product.cpu.abi`, and `ro.product.cpu.abilist` is not used at all (at least on arch detection).

osm0sis:
Yeah the reason I do getprop there is since /system is mounted, so it should be system's getprop and values hopefully, but then I also fallback to default.prop, which is recovery's, and may be wrong.

ipdev99:
> Thanks for noticing and figuring this out. But it is weird, `abilist` is the recommended field. `abi` and `abi2` should be obsolete by now :-/
> 
> On which ROMs does this happen, do we have some examples of the prop-files?
> 
> I am a bit hesitant about this change, because I fear we could break other ROMs (and e.g. emulators).
> 
> If possible I'd prefer a more surgical method to detect these broken ROM prop files and deal with it properly.

I am using an Android 12 rom on my Poco F1 (beryllium).
_Unofficial AospExtended v9.0 rom._
_Unofficial OpenGApps 12 builds._
_Official TWRP 3.5.2_9-0_

Since the installer script finds the device as arm, I was able to install the arm version of OpenGApps 12.
Reversing the abi check allowed me to install the arm64 version.

Part of figuring this out, I ran a [script](https://github.com/ipdev99/example/blob/opga/find_abi.sh) that used the grep command to find `cpu.abi` values in the props files.

[beryllium_abi_list.log](https://github.com/ipdev99/example/blob/opga/beryllium_abi_list.log)

The prop files are in the [beryllium](https://github.com/ipdev99/example/tree/opga/beryllium) directory.
_default.prop is a symlink to prop.default so I only included prop.default_

The [check_abi](https://github.com/ipdev99/example/blob/opga/check_abi.sh) script is what I have been using to test the return value of `$device_architecture`
_(I switch the `-n` to `-z` to move down the list for testing each one.)_

_Note:
I am booted into TWRP and mount all partitions before running the scripts._

ipdev99:
I originally asked for opinions in the xda discussion thread since I am still trying to work it out.

As I noted in my [xda post](https://forum.xda-developers.com/t/gapps-daily-open-gapps-for-android-all-android-versions-devices.3098071/post-85951207).
_Looking for opinions on reversing the order of (or improving) the abi check._

I am not that fond of just reversing it.

I am also looking for thoughts about including (using) other values.

`ro.vendor.product.cpu.abilist` `ro.odm.product.cpu.abilist` `ro.system.product.cpu.abilist`



I am trying to keep this as clean as possible and make sure it is backwards compatible.
_Hopefully a bit future proof also._

This is my current idea.
[WIP abi check](https://github.com/ipdev99/opengapps/commit/a3015a5c7ac5d42769307a0e8e2be3644b57b94b)

- Set `device_architecture` using `ro.product.cpu.abilist` if it matches `ro.system.product.cpu.abilist`.
  - If both are empty the condition will still equal.
- If empty value, try to set it from another prop value.
  - If exists, use `ro.system.product.cpu.abilist`
  - If exists, use `ro.vendor.product.cpu.abilist`
  - Fall back to `ro.product.cpu.abi` if it still exists.
  - Give up and set it to what ever `ro.product.cpu.abilist` returns.

mfonville:
@ipdev99 I think your approach is the right way to do it. If you are done with putting that in code, that would be very welcome.
We also would need some (real world) testing before we can merge it.

nezorflame:
Agreed.
@ipdev99 wanna make a PR? I'll close mine then.

ipdev99:
@nezorflame
Made a PR https://github.com/opengapps/opengapps/pull/944

I know there are a few user that will be happy to try your test builds.
_I think there was at least  one in the xda thread and one or two in your TG channel that are waiting for this fix._

nezorflame:
@ipdev99 thanks! Yeah, will make a build, let's switch to your PR.

PR #943, opened 2021-11-16
[WIP] Add permission check script

nezorflame:
Added Python script (modified version of [MindTheGApps one](https://gitlab.com/MindTheGapps/vendor_gapps/-/blob/sigma/cicd/verify-permissions.py)) to check missing privapp permissions.

Also modified [add_sourceapp.sh](https://github.com/opengapps/opengapps/pull/943/files#diff-e050316f39f2890ac011b0c838420c567c781f32f637b5523b4f4e1fd9177070R162) so that it also checks missing permissions for the added APKs.

### DIFF
--- a/add_sourceapp.sh
+++ b/add_sourceapp.sh
@@ -158,6 +158,8 @@ $notinzip";;
   if echo "$architectures" | grep -q "all"; then #no space (single entry)
     installapk "all"
   fi
+
+  checkpermissions "$apk"
 }
 
 addlib() {
@@ -181,6 +183,12 @@ addlib() {
     (*)           install -D "$lib" "$SOURCES/$architecture/$prefix$libfolder/$REPLY/$libname"
                   echo "SUCCESS: Added $libname to $architecture/$prefix$libfolder/$REPLY/";;
   esac
+}
+
+checkpermissions() {
+  apk="$1"
+  echo "Checking if any extra privapp-permissions are needed..."
+  python scripts/verify-permissions.py "$apk"
 }
 
 for argument in "$@"; do
--- a/add_sourceapp.sh
+++ b/add_sourceapp.sh
@@ -28,7 +28,7 @@ BETA=""
 STUB=""
 
 # Check tools
-checktools aapt file coreutils java jarsigner keytool openssl unzip
+checktools aapt file coreutils java jarsigner keytool openssl python3 unzip
 
 installapk() {
   architecture="$1"
@@ -139,6 +139,14 @@ $notinzip";;
   fi
   echo "APK is complete, certificate is valid and signed by Google"
 
+  checkpermissions "$apk"
+  checked="$?"
+  if [ "$checked" != 0 ]; then
+    echo "ERROR: $missingperms"
+    return 1
+  fi
+  echo "No additional privapp-permissions are needed"
+
   #We manually check for each of our set of supported architectures
   #We assume NO universal packages for 32vs64 bit, so start with the 'highest' architectures first, if it matches one of those, we will NOT add it to a lower architecture
   if { echo "$architectures" | grep -q "armeabi" && ! echo "$architectures" | grep -q "arm64"; } ||\
@@ -158,8 +166,6 @@ $notinzip";;
   if echo "$architectures" | grep -q "all"; then #no space (single entry)
     installapk "all"
   fi
-
-  checkpermissions "$apk"
 }
 
 addlib() {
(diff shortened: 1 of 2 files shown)


add_sourceapp.sh
```diff
@@ -183,6 +185,12 @@ addlib() {
   esac
 }
 
+checkpermissions() {
+  apk="$1"
+  echo "Checking if any extra privapp-permissions are needed..."
+  python scripts/verify-permissions.py "$apk"
```
mfonville, line 191: This should be `python3`

mfonville: Line 31 with checktools also should check if `python3` is available

scripts/verify-permissions.py
```diff
@@ -0,0 +1,143 @@
...
+# This script helps to fix the private permission issues caused by updated APKs.
+# See https://gitlab.com/MindTheGapps/vendor_gapps/-/blob/sigma/cicd/verify-permissions.py
+# for the original file.
+
+import errno
+from glob import glob
+import os
+import subprocess
+import sys
+from xml.etree import ElementTree
```
mfonville, line 25: Also do a check if these imported libraries are available in our checktools (or other graceful dealing with dependencies)

scripts/verify-permissions.py
```diff
@@ -0,0 +1,143 @@
...
+    exit(errno.ENOPKG)
+
+# Change working directory to the location of this script
+# This fixes relative path references when calling this script from
+# outside of the directory containing it
+os.chdir(sys.path[0])
+
+# Definitions for privileged permissions
+ANDROID_MANIFEST_XML = \
+    'https://raw.githubusercontent.com/LineageOS/android_frameworks_base/lineage-19.0/core/res/AndroidManifest.xml'
```
mfonville, line 48: Can we host our own file for this?

mfonville: And rather also a local copy in our repository than requesting a file from the internet every time we add an apk.

scripts/verify-permissions.py
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/python3
+#
+# Copyright (C) 2021 Paul Keith <javelinanddart@gmail.com>
+#
```
mfonville, line 4: If you made any changes, also add your own name

mfonville (commented):
Thanks @nezorflame !

I have some small points of feedback, but already looks very good

add_sourceapp.sh
NicholasBuse: Actually, I do have a comment! Can we return the exit code from the python script here (and not install the apk above if unsuccessful)?

scripts/verify-permissions.py
nezorflame: You mean instead of the original author's name or next to it? I've placed my name lower on line 14.

add_sourceapp.sh
nezorflame: done and done

add_sourceapp.sh
```diff
@@ -158,6 +158,8 @@ $notinzip";;
   if echo "$architectures" | grep -q "all"; then #no space (single entry)
     installapk "all"
   fi
+
+  checkpermissions "$apk"
```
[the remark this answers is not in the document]
nezorflame: done

add_sourceapp.sh
```diff
@@ -188,7 +194,10 @@ addlib() {
 checkpermissions() {
   apk="$1"
   echo "Checking if any extra privapp-permissions are needed..."
-  python scripts/verify-permissions.py "$apk"
+  missingperms=$(python3 scripts/verify-permissions.py "$apk" 2>&1 >/dev/null)
```
mfonville, line 197: does the python script give a return code? That is probably easy to check for than testing if it returns a string.

add_sourceapp.sh
```diff
@@ -139,6 +139,14 @@ $notinzip";;
   fi
   echo "APK is complete, certificate is valid and signed by Google"
 
+  checkpermissions "$apk"
+  checked="$?"
```
mfonville, line 143: Checking the return-code here is not very easy to read/understand.

I think it would be better to handle both the checking and any errors in the checkpermission() function itself.

scripts/verify-permissions.py
mfonville: IMHO you should move line 14 to here

lozanosteven79-afk (approved):

PR #960, opened 2022-05-11
Android Auto

ipdev99:
Change in Android 12 and/or 12L.
_Not sure if this is required for Android 12._
`SYSTEM_AUTOMOTIVE_PROJECTION` - [android.googlesource.com](https://android.googlesource.com/platform/frameworks/base/+/8240dbd571cb86b04ed0fe252eeb381971b7a54d^!/)

First time I made an overlay. :upside_down_face:

Not sure if I set `android:priority` correctly.
_Got a little confused reading about setting it._

My WIP overlay - [GitLab](https://gitlab.opengapps.org/ip/all/-/commit/69c2bfdb938577eb842fbb82a3b9760ea270827d)

Cheers all. :cowboy_hat_face:

### DIFF
--- a/scripts/inc.buildtarget.sh
+++ b/scripts/inc.buildtarget.sh
@@ -211,7 +211,11 @@ get_package_info(){
     webviewgoogle)            packagetype="GApps"; packagename="com.google.android.webview"; packagetarget="app/WebViewGoogle"; packagegappsremove="$webviewgappsremove";;
     trichromelibrary)         packagetype="GApps"; packagename="com.google.android.trichromelibrary"; packagetarget="app/TrichromeLibrary";;
     webviewstub)              packagetype="GApps"; packagename="com.google.android.webview.stub"; packagetarget="app/WebViewStub";;
-    gearheadstub)             packagetype="GApps"; packagename="com.google.android.projection.gearhead.stub"; packagetarget="priv-app/AndroidAutoPrebuiltStub";;
+    gearheadstub)             packagetype="GApps"; packagename="com.google.android.projection.gearhead.stub"; packagetarget="priv-app/AndroidAutoPrebuiltStub"
+                              if [ "$API" -ge "31" ]; then  # Add an overlay for Android 12.0 +
+                                packagefiles="product/overlay/AndroidAutoOverlay.apk"
+                                packagegappsremove="product/overlay/AndroidAutoOverlay.apk vendor/overlay/AndroidAutoOverlay.apk"
+                              fi;;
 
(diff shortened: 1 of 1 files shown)


mfonville:
I also have no experience with overlays.

Did you test it?

@osm0sis what is your take on overlays?

osm0sis:
I know we use several already, IIRC to avoid Pixel Launcher crashes, etc. Not sure if there's a preference for how they're built/signed or anything, perhaps @nezorflame or @NicholasBuse know more about the overlays in use already.

ipdev99:
Tested on arm and arm64.
Pixel 5a, official Lineage 19.1
Nexus 7 [2013] (deb) LTE/Mobile, unofficial lineage 19.1
Aftermarket head unit.

ipdev99:
After reading about overlays [source.android](https://source.android.com/devices/architecture/rros).
I used nezorflame's guide [gitlab.opengapps](https://gitlab.opengapps.org/opengapps/all/-/blob/master/product/overlay/README.md) to build the apk.

I originally sent a PM to nezorflame on xda about it since I could not create an issue/PR on opengapps.gitlab at the time.

He told me to create a PR without the apk, he would build it and sign it with the official key.

ipdev99:
On the topic of overlays.
There is another change in Android that I missed or have not found reference to yet.

I get a warning in log files about the overlays.
`targets an SDK below the required SDK level of system overlays (32). This will become an install error in a future release`
[Segment from logcat](https://bpa.st/S2FA) for example.

So it looks like we will need to change the way overlays are handled.
_Prebuild overlays per SDK level and/or build them inline with the package builds?_

That will be the topic of a different issue.:upside_down_face:

josephcsible:
Since the target SDK and minimum SDK are separate, couldn't we just increase the former and leave the latter alone, and then still be able to use the same APK in every build going forward?

ipdev99:
Not sure about backwards compatibility since I have not found where/when this change comes from.
_Not sure if it will cause an error using an overlay targeting a higher level than the Android build level._

Going forward, each overlay would still need to be rebuilt to target a newer version every time a new Android level is released.
_Even if all the overlays are rebuilt to target SDK32.
They will need to rebuilt again to target SDK33 when Android 13 is released.
They might also need to be updated to comply with the new target SDK level._

Why this is a topic of discussion for it's own issue.
> That will be the topic of a different issue. :upside_down_face:

:cowboy_hat_face:

mfonville:
ok, looking forward to a research and solution for that :-)

I will merge this, thanks!

PR #961, opened 2022-05-16
Add priv-app/MotCamera2 to the removal list

josephcsible:
This is where the stock camera lives on LineageOS on a Moto X4 "payton".


mfonville:
Thanks!

PR #974, opened 2022-12-05
Add Persian Language

RminFX:
Add Persian Language

Fixes #.

Changes:
-
-
-


mfonville:
Hi @RminFX,

thank you for your PR, but I think it is better to create the PR at the Aroma-project itself: https://github.com/amarullz/AROMA-Installer

We are only re-using their work for our Aroma-package.

PR #979, opened 2023-06-14
installer: update mount logic + make comments uniform

osm0sis:
- update apex mounting support for Android 11+
- check before mounts, fix persist mount on Virtual A/B
- fix apex mounts on fastboot boot'd TWRP
- support Android 12 .capex (compressed apex) files
- fix some (bad) OEM prop files including entries twice, use last entry like live system
- add useful mount debugging stderr for logs
- don't unmount any existing apex at start of flash
- support AMLogic devices where A-only sets slot=normal
- correct build.prop locations for getprop replacements
- support /metadata mount as a safer alternative to /persist
- refactor mount_all and umount_all
- use OLD_PATH for exceptions where busybox causes conflicts
- abort apex loop mounts early if failing
- work around loop mounts on newer Lineage-based recoveries by changing shell context

### DIFF
--- a/scripts/inc.installer.sh
+++ b/scripts/inc.installer.sh
@@ -80,7 +80,7 @@ makeupdatebinary() {
 #    zip exception de-facto LGPLv3 licensed.
 #
 export OPENGAZIP="$3"
-export OUTFD="/proc/self/fd/$2"
+export OUTFD="$2"
 export TMP="/tmp"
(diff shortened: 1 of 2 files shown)


osm0sis:
Yes, it's the latest in AnyKernel3, running on devices without issue for the last 3 weeks. Seems stable.

